PR #2, opened 2017-03-31
fix(server-sent demo): fix PHP notice

pa-de-solminihac:
just fixing a notice - and sneakily removing the php closing tag ;)


chrisdavidmills:
looks good - thanks for the fix!

PR #10, opened 2017-10-22
Create examples of Streams API

ksm2:
Hey, I created two examples of the Streams API:
* **Simple Stream Pump**. This example shows how to consume a ReadableStream and pass its data to another.
* **Grayscale a PNG**. This example shows how a ReadableStream of a PNG can be turned into grayscale.


chrisdavidmills:
Thanks Konstantin!

PR #12, opened 2017-10-24
[Streams] Fix PNG chunk CRC and stream end

ksm2:
Now the example also works in FF and pngchecker is reporting no errors anymore.


chrisdavidmills:
Fabulous - thanks!

PR #14, opened 2017-12-24
Fix `insertAdjacentElement` example

ufocoder:
If user will not select an div element and will click on any button then will be an error like the followings:
```Uncaught TypeError: Cannot read property 'insertAdjacentElement' of undefined```


chrisdavidmills:
Cool, thanks for the fix!

PR #15, opened 2018-01-17
Corrected display of JS-object in HTML by using JSON.stringify

MaxSchumacher:
without JSON stringify, the display is [Object Storage] in the HTML display


chrisdavidmills:
Ah, good catch - thanks!

PR #16, opened 2018-02-13
Payment Request examples

poshaughnessy:
Hi. As discussed here: https://github.com/w3c/payment-request-info/issues/4

This adds Payment Request examples, based on the code snippets from https://github.com/w3c/payment-request-info/wiki/CodeExamples

I hope it makes sense to start this PR review process, so people can add feedback here, although I have a few questions remaining related to Payment Handlers, please: 

* Sorry I am unfamiliar: will BobPay (the native app and/or the web app) need to whitelist the MDN domain to allow the Payment Request? (I've noticed the `supported_origins` [here](https://rsolomakhin.github.io/bobpay/payment-manifest.json), which is linked to [here](https://github.com/madmath/payment-request-show/blob/d02c2d61703aa706aa1bdc7355fb80474a8acd8d/bobpay/app.js)?)
* ~~(Possibly related...) I have tested that the 'Recommend a Payment App' demo works when I have the BobPay native Android app installed, but when I only have the web app version of the Payment Handler installed, it gives me the 'NOT_SUPPORTED_ERR` and redirects to the download page. I wonder what is missing to allow the web app to work too?~~ Sorry, scratch that. It's working in Chrome Canary.
* For the 'Pre-authorize Transaction' demo: are there any Payment Handler demos out there which implement this? Or we need to build one?

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ Code examples that accompany various MDN DOM and Web API documentation pages.
 
 * The "media" directory contains examples and demos showing how to use HTML and DOM [media elements and APIs](https://developer.mozilla.org/en-US/docs/Web/Media).
 
+* The "payment-request" directory contains examples of the Payment Request API.
+
 * The "performance-apis" directory is for examples and demos of the [Web Performance APIs](https://www.w3.org/wiki/Web_Performance/Publications).
 
 * The "pointerevents" directory is for examples and demos of the [Pointer Events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events) standard.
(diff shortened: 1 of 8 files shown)


rsolomakhin:
> will BobPay (the native app and/or the web app) need to whitelist the MDN domain to allow the Payment Request? (I've noticed the supported_origins here, which is linked to here?)

No need to whitelist MDN. MDN is a merchant. Our app allows all merchants to use it. The `supported_origins` in the manifest indicates which domains can have _payment apps_ (not merchants) that handle the BobPay payment method.

> For the 'Pre-authorize Transaction' demo: are there any Payment Handler demos out there which implement this? Or we need to build one?

I'm not aware of any Payment Handler demos that pre-authorize transactions. This would happen in the `"canmakepayment"` event handler. You can see one at the top of https://bobpay.xyz/pay/sw-bobpay.js, but it does not perform any complex logic and rubber stamps all transactions:

```javascript
self.addEventListener('canmakepayment', function(e) {
  e.respondWith(true);
});
```

PR #18, opened 2018-03-29
Update stream examples to current spec

ksm2:
The spec now includes TransformStreams and encourages the user to use async/await code with while-loops instead of recursive function calls. I updated my examples based on feedback I got for [a PR I filed against the spec repository to add demos](https://github.com/whatwg/streams/pull/911). The examples are tested on Chrome with the _Experimental Web Platform features_ flag (chrome://flags/#enable-experimental-web-platform-features) activated.

See PR: https://github.com/whatwg/streams/pull/911

### DIFF
--- a/streams/grayscale-png/index.html
+++ b/streams/grayscale-png/index.html
@@ -22,11 +22,15 @@
   // Retrieve its body as ReadableStream
   .then(response => response.body)
   // Create a gray-scaled PNG stream out of the original
-  .then(body => new Response(new ReadableStream(new GrayscalePNGSource(body))))
-  .then(newResponse => newResponse.blob())
+  .then(rs => rs.pipeThrough(new TransformStream(new GrayscalePNGTransformer())))
+  // Create a new response out of the stream
+  .then(rs => new Response(rs))
+  // Create an object URL for the response
+  .then(response => response.blob())
   .then(blob => URL.createObjectURL(blob))
+  // Update image
   .then(url => image.src = url)
-  .catch(err => console.error(err));
+  .catch(console.error);
 
 </script>
 </body>
--- a/streams/grayscale-png/png-lib.js
+++ b/streams/grayscale-png/png-lib.js
@@ -19,181 +19,154 @@ function paeth(a, b, c) {
 
 
 /**
- * A source of a ReadableStream to create a grayscale PNG.
+ * A transformer of a TransformStream to create a grayscale PNG.
  */
-class GrayscalePNGSource {
-  /**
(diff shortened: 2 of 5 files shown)


chrisdavidmills:
Hi @ksm2 , nice to hear from you! Apologies for taking a it of time to respond back.

This looks like some amazing work. before I merge this, I had some questions:

1. Does this stuff work yet in Firefox when enabled?
2. Is it going to remove examples that demonstrate specific Streaming features that we refer to in the Streams docs on MDN? If so, it might be better to include the updated examples in separate directories instead of overwriting them.

Let me know your thoughts, and then we'll continue this onwards. Thanks.

ksm2:
Hi @chrisdavidmills , no problem!

1. Unfortunately not. Firefox (and I even checked the Nightly 61) has no support for `WritableStream` or `TransformStream` yet. But there is a [polyfill](https://github.com/creatorrr/web-streams-polyfill/blob/master/dist/polyfill.js).
2. AFAIK, the grayscale PNG example is only referenced on the [Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) page. I fixed a comment in the PNG transform stream example which is also referenced in the [pipeThrough](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/pipeThrough) documentation.

Maybe we should add the new classes `TransformStream`, `TransformStreamDefaultController`, etc. also to the MDN. They have a slightly different interface then the `ReadableStream`. Then, I could write a simple `Uint8ArrayToLinesOfTextTransform` example for TransformStreams and their Controllers. What do you think of it?

chrisdavidmills:
I didn't think so; I know that work has stalled on our side somewhat. Still doesn't mean that we shouldn't make the updates.

I will have to try to find some time to add the new pages. Can you write the new example anyway? It would help me in understanding the new interfaces ;-)

chrisdavidmills:
Looking at the spec, I think the only new stuff is TransformStream and TransformStreamDefaultController. This shouldn't be too much work to add. An example would still help though ;-)

ksm2:
Hey Chris, I created the example for you! Take a look: https://gist.github.com/ksm2/76bd39ce41935c8299a13436ae430964

chrisdavidmills:
That is great - thanks! I will merge this PR now, and then find some time to get through the pages and check example listings very soon.

Can your open up a new PR to add the new example?

chrisdavidmills:
Ah, I already merged it, d'oh. Losing my mind here.

PR #19, opened 2018-04-26
Add example to transform binary chunks to strings

ksm2:
Hey @chrisdavidmills,
here is the binary chunks to strings example now as a PR!
I replaced _The Raven_ with Goethe’s _Faust, Part One_ as the former was not long enough to require several 64 KiB chunks – and _Faust, Part One_ was one of the longest works I remembered from school. :smile: 

FYI, I added a link to a non-existing page on the MDN: https://developer.mozilla.org/en-US/docs/Web/API/TransformStream. I guess you will put the new page there, right?


chrisdavidmills:
Thanks @ksm2, this looks great. Yes, I will add the TransformStream docs there. I am probably going to be doing this during our next sprint, so in about 2-3 weeks time.

PR #20, opened 2018-05-10
Added server-sent-events/README.md

duzun:
I've spent some time trying to make this example work with no luck, and in the end, I've found that `fastcgi_buffering` was causing troubles.
I think it is worth mentioning this small fact.

### DIFF
--- a/server-sent-events/README.md
+++ b/server-sent-events/README.md
@@ -1,0 +1,3 @@
+## Note about running this example through **NGINX**:
+
+By default `fastcgi_buffering` is ON, thus you have to add `fastcgi_buffering off;` to the appropriate `location` block for this example to work.


server-sent-events/README.md
```diff
@@ -0,0 +1,3 @@
+Note about runing this example through **NGINX**:
```
chrisdavidmills: Thanks for the addition! Couple of change requests:

* Typo - runing -> running
* Can you turn this line into a heading?

duzun:
@chrisdavidmills done.

Is it worth adding `while(!connection_aborted()) ...` for this example, what do you think?

chrisdavidmills:
Thanks! I'll merge this now. I think it might be worth considering adding a conditional check, yes. Do you fancy handling that in a separate PR?

duzun:
Cool!

PR #21, opened 2018-05-16
server-sent-events: Break the while(1) loop on connection_aborted()

duzun:
Break the while(1) loop on connection_aborted() so that PHP process terminates as soon as it is not of use.


chrisdavidmills:
Looks good, thanks!

Issue #25, opened 2018-07-06
Channel Messaging Examples are confusing at best.


Kaiido:
I fail to see the point of [channel-messaging-multimessage](https://github.com/mdn/dom-examples/tree/master/channel-messaging-multimessage) example...

Why is a new `MessageChannel` created every time, instead of simply passing the new messages in the MessagePort?  
That makes the use of `WindowObject.postMessage` and `MessagePort.postMessage` really unclear: it sounds like either we can postMessage to only a single port, or that MessageChannel is a *single-shot* API, this later speculation being comforted by weird wording in MDN's [Using channel messaging](https://developer.mozilla.org/en-US/docs/Web/API/Channel_Messaging_API/Using_channel_messaging#Use) article:

> Message channels on the other hand can provide a secure channel that **a single data item** can be sent down, from one browsing context to another, **after which the channel is closed**. 
<!--><sup>[emphasize mines]</sup>

That surely can be heard as channel messaging **is** "single-shot", and it actually have been at least by the asker of [this question on stackoverflow](https://stackoverflow.com/questions/51201120/using-messagechannel-bidirectionally-for-multiple-messages-between-page-and-if/).

So except if it is really a "single-shot" API and that **I** got it all wrong, (in this case all implementations would have got it wrong too)  I think *multimessage* is actually a *basic usage*, that it should be rewritten to show a basic usage (i.e reusing the MessagePorts), and that what is currently called *"basic"* should be renamed *single-message*.

chrisdavidmills:
Hrm, I think you have got a good point here. This is rather confusing. Would you be willing to help us update the examples and the text? I'd be happy to review a pull request to improve these demos.

ollelauribostrom:
@chrisdavidmills @Kaiido 

I've made some updates to the docs which hopefully makes things less confusing (and that uses the updated example provided in my PR). What do you think? Would love some feedback on improvements etc. 

Updated pages:
- https://developer.mozilla.org/en-US/docs/Web/API/Channel_Messaging_API/Using_channel_messaging
- https://developer.mozilla.org/en-US/docs/Web/API/MessageChannel
- https://developer.mozilla.org/en-US/docs/Web/API/MessagePort

chrisdavidmills:
Looks good; I've given the pages a copy edit.

PR #27, opened 2018-09-11
Improving channel messaging examples

ollelauribostrom:
This addresses #25.

- The `channel-messaging-basic` has been cleaned up a bit
- The `channel-messaging-multimessage` does no longer create a new `MessageChannel` every time. Instead a single `MessageChannel` is created, and one of its ports is initially transfered to the iframe using `iframe.contentWindow.postMessage`. All succeeding messages sent between the main frame and the iframe are sent using the ports.

I would be happy to update the text as well, assuming I got this right.


chrisdavidmills:
These changes look good, thanks!

If you could update the associated docs too, that'd be great.

PR #28, opened 2018-09-20
Use switch, srcObject, return response.blob. Attempt to make video download not one time.

ZaneHannanAU:
Essentially increase usability and not give errors in console.

Currently; the abort does not trigger after fetch starts and will error in console.

### DIFF
--- a/abort-api/index.html
+++ b/abort-api/index.html
@@ -45,8 +45,7 @@
   <script>
     var url = 'sintel.mp4';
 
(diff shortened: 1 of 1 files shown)


ZaneHannanAU:
Ah… oops yeah…

ZaneHannanAU:
... fixed; thank you.

Not going to try to use mediastream... that's a whole other can of worms.

PR #32, opened 2018-11-28
[do not merge yet] Add crypto examples

wbamberg:
This PR adds some examples for the [`sign()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign) and [`verify()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign) methods of the Web Crypto API.

There are 4 examples, one for each supported algorithm. This will also cover part of the [`generateKey()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign) method, because we're generating the key pairs to use.

I've structured it to keep the implementations for each algorithm as self-contained as possible, so they can be understood in isolation.

We should get a review from the security team before merging, but it would be good to get some feedback at this point anyway, as I'm intending to use this template for the other crypto methods, like [`encrypt()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/encrypt)/[`decrypt()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/decrypt) and [`digest()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest).


wbamberg:
ok, let's try this again.

PR #33, opened 2018-11-28
Add sign/verify examples for Web Crypto

wbamberg:
This PR adds some examples for the [`sign()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign) and [`verify()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign) methods of the Web Crypto API.

There are 4 examples, one for each supported algorithm. This will also cover part of the [`generateKey()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign) method, because we're generating the key pairs to use for the 4 algorithms included.

I've structured it to keep the implementations for each algorithm as self-contained as possible, so they can be understood in isolation.

We should get a review from the security team before merging, but it would be good to get some feedback at this point anyway, as I'm intending to use this template for the other crypto methods, like [`encrypt()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/encrypt)/[`decrypt()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/decrypt) and [`digest()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest).

### DIFF
--- a/web-crypto/sign-verify/rsa-pss.js
+++ b/web-crypto/sign-verify/rsa-pss.js
@@ -58,11 +58,7 @@
       encoded
     );
 
-    if (result) {
-      signatureValue.classList.add("valid");
-    } else {
-      signatureValue.classList.add("invalid");
-    }
+    signatureValue.classList.add(result ? "valid" : "invalid");
   }
 
   /*
--- a/web-crypto/sign-verify/rsa-pss.js
+++ b/web-crypto/sign-verify/rsa-pss.js
@@ -68,6 +68,7 @@
   window.crypto.subtle.generateKey(
     {
       name: "RSA-PSS",
+      // Consider using a 4096-bit key for systems that require long-term security
       modulusLength: 2048,
       publicExponent: new Uint8Array([1, 0, 1]),
       hash: "SHA-256",
--- a/web-crypto/sign-verify/rsa-pss.js
+++ b/web-crypto/sign-verify/rsa-pss.js
@@ -68,7 +68,7 @@
   window.crypto.subtle.generateKey(
     {
       name: "RSA-PSS",
-      modulusLength: 2048,
+      modulusLength: 4096,
       publicExponent: new Uint8Array([1, 0, 1]),
       hash: "SHA-256",
     },
--- a/web-crypto/sign-verify/rsa-pss.js
+++ b/web-crypto/sign-verify/rsa-pss.js
@@ -1,4 +1,4 @@
-(function() {
+(() => {
 
   /*
   Store the calculated signature here, so we can verify it later.
--- a/web-crypto/sign-verify/rsa-pss.js
+++ b/web-crypto/sign-verify/rsa-pss.js
@@ -1,0 +1,95 @@
+(function() {
+
+  /*
+  Store the calculated signature here, so we can verify it later.
+  */
+  let signature;
+
+  /*
+  Fetch the contents of the "message" textbox, and encode it
+  in a form we can use for sign operation.
+  */
+  function getMessageEncoding() {
+    const messageBox = document.querySelector(".rsa-pss #message");
+    let message = messageBox.value;
+    let enc = new TextEncoder();
+    return enc.encode(message);
+  }
+
+  /*
+  Get the encoded message-to-sign, sign it and display a representation
+  of the first part of it in the "signature" element.
+  */
+  async function signMessage(privateKey) {
+    const signatureValue = document.querySelector(".rsa-pss .signature-value");
+    signatureValue.classList.remove("valid");
+    signatureValue.classList.remove("invalid");
+
+    let encoded = getMessageEncoding();
+    signature = await window.crypto.subtle.sign(
+      {
+        name: "RSA-PSS",
+        saltLength: 128,
+      },
+      privateKey,
+      encoded
+    );
+
+    let buffer = new Uint8Array(signature, 0, 10);
+    signatureValue.textContent = buffer;
+  }
+
+  /*
+  Fetch the encoded message-to-sign and verify it against the stored signature.
+  * If it checks out, set the "valid" class on the signature.
+  * Otherwise set the "invalid" class.
+  */
+  async function verifyMessage(publicKey) {
+    const signatureValue = document.querySelector(".rsa-pss .signature-value");
+    signatureValue.classList.remove("valid");
+    signatureValue.classList.remove("invalid");
+
+    let encoded = getMessageEncoding();
+    let result = await window.crypto.subtle.verify(
+      {
+        name: "RSA-PSS",
+        saltLength: 128,
+      },
+      publicKey,
+      signature,
+      encoded
+    );
+
+    if (result) {
+      signatureValue.classList.add("valid");
+    } else {
+      signatureValue.classList.add("invalid");
+    }
+  }
+
+  /*
+  Generate a sign/verify key, then set up event listeners
+  on the "Sign" and "Verify" buttons.
+  */
+  window.crypto.subtle.generateKey(
+    {
+      name: "RSA-PSS",
+      modulusLength: 2048,
+      publicExponent: new Uint8Array([1, 0, 1]),
+      hash: "SHA-256",
+    },
+    true,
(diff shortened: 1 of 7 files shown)


wbamberg:
@April, here's an attempt to add some examples for the [`sign()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign) and [`verify()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign) methods of the Web Crypto API.

Thanks for offering to take a look!

web-crypto/sign-verify/rsa-pss.js
```diff
@@ -0,0 +1,95 @@
...
+  * If it checks out, set the "valid" class on the signature.
+  * Otherwise set the "invalid" class.
+  */
+  async function verifyMessage(publicKey) {
+    const signatureValue = document.querySelector(".rsa-pss .signature-value");
+    signatureValue.classList.remove("valid");
+    signatureValue.classList.remove("invalid");
+
+    let encoded = getMessageEncoding();
+    let result = await window.crypto.subtle.verify(
```
april: The current documentation for `subtle.crypto.verify()` only mentions `HMAC`, `RSASSA-PKCS1-v1_5` and `ECDSA` as options. It also lists the algorithm as a `DOMString`, not an object. I'm not sure if that's just a simple tweak or if there is intended to be an overhaul of the function documentation as well.

web-crypto/sign-verify/rsa-pss.js
```diff
@@ -0,0 +1,95 @@
...
+
+  /*
+  Fetch the encoded message-to-sign and verify it against the stored signature.
+  * If it checks out, set the "valid" class on the signature.
+  * Otherwise set the "invalid" class.
+  */
+  async function verifyMessage(publicKey) {
+    const signatureValue = document.querySelector(".rsa-pss .signature-value");
+    signatureValue.classList.remove("valid");
+    signatureValue.classList.remove("invalid");
```
april: `classList.remove()` should take multiple arguments, which cleans this up a little bit.

web-crypto/sign-verify/rsa-pss.js
wbamberg: Yes, those are errors in the docs:

* RSA-PSS is also supported (https://www.w3.org/TR/WebCryptoAPI/#rsa-pss).

* Wherever `algorithm` is an object I want to have a separate page describing the object (i.e. `RsaPssParams`, `EcdsaParams`, that's what I'm hoping to do in https://github.com/mdn/sprints/issues/667.

april:
>     signatureValue.classList.remove("valid");
>     signatureValue.classList.remove("invalid");

`classList.remove()` should take multiple arguments, which cleans this up a little bit.

>     let result = await window.crypto.subtle.verify(
>       {
>         name: "RSA-PSS",
>         saltLength: 32,
>       },
>       publicKey,
>       signature,
>       encoded
>     );

The current documentation for `subtle.crypto.verify()` only mentions `HMAC`, `RSASSA-PKCS1-v1_5` and `ECDSA` as options. It also lists the algorithm as a `DOMString`, not an object. I'm not sure if that's just a simple tweak or if there is intended to be an overhaul of the function documentation as well.

april:
A lot of the examples use:

>     if (result) {
>       signatureValue.classList.add("valid");
>     } else {
>       signatureValue.classList.add("invalid");
>     }

In the interests of brevity (while still maintainable legibility) I might recommend:
```javascript
signatureValue.classList.add(result ? "valid" : "invalid");
```

I've also noticed that the code has both `async`/`await` as well as `Promise.then()`, as well as a mix of `function()` and arrow syntax. Not sure if MDN has any specific requirements, but it might be nice to stick to a single syntax?

```javascript
generateSignVerifyKey = window.crypto.subtle.generateKey(
```

It doesn't appear you actually use this value anywhere.

april:
>   window.crypto.subtle.generateKey(
>     {
>       name: "HMAC",
>       hash: {name: "SHA-384"}
>     },
>     true,
>     ["sign", "verify"]

For HMAC, I would probably suggest using either SHA-256 or SHA-512, with a nod towards SHA-512 as it is more secure and faster on 64-bit machines.

april:
```
window.crypto.subtle.generateKey(
    {
      name: "RSASSA-PKCS1-v1_5",
      modulusLength: 2048,
      publicExponent: new Uint8Array([1, 0, 1]),
      hash: "SHA-256",
    },
```

A 2048-bit modulus is probably fine for most use cases, provided that the key is not used for an extended period of time. If we don't want to get into the nuances, it probably wouldn't hurt to simply use a 4096-bit modulus with SHA-256. That's what I would personally recommend myself.

wbamberg:
Thanks @april for your comments! I've addressed most but not all of them. Specifically, I've done:

* use multiple arguments for classList.remove
* use ternary operator to set the validity classes
* SHA-512 for HMAC
* remove unnecessary variable assignment from  `  window.crypto.subtle.generateKey`
* switched to arrow functions for the top level IIFE. I didn't see anywhere else that could use them.

I tried switching to 4096-bit RSA, but on my computer it took several seconds before the page was usable. We could disable the buttons until then, but even then it would seem quite weird.

I kept the `Promise.then` at the top level, as it seemed more natural to me ("generate the key, then set up the listeners"). AFAIK we don't have a standard about this. But this is quite subjective I know.

april:
Huh, weird about RSA 4096. I didn't experience any particular issues on my laptop, but I guess systems might vary quite a bit.

That one is no big deal either way, and we can always make a note to consult relevant documentation on what key size to use for the given use case.

wbamberg:
> That one is no big deal either way, and we can always make a note to consult relevant documentation on what key size to use for the given use case.

Yes, maybe worth adding a comment like "Consider using 4096 here" ?

april:
I would probably say something like

```javascript
// Consider using a 4096-bit key for systems that require long-term security
```

And maybe link to keylength.com on the `generateKey()` and `deriveKey()` pages?

wbamberg:
> I would probably say something like

Done!

> And maybe link to keylength.com on the generateKey() page?

Yes, good idea.

wbamberg:
@chrisdavidmills , thanks for your comments too.

> I'd love to see what the code looks like that is doing the signing and verifying, while I am playing with these. Could you dump the contents of the relevant JS file in a box on the bottom left of the UI, while each box on the left is being played with? Or do you think that'd get too confusing?

I'm not sure how to implement this sensibly, without duplicating the code. I would expect someone who wanted to see the code would just clone it or browse it on GitHub. I do want to keep examples as simple as possible, focusing on the usage of the APIs and minimising any extraneous UI as far as possible.

(Another way to say this is: in my mind there are three interlocking pieces here: the example page, which shows you what it looks like, the source code, which shows you how it works, and the docs page. I think keeping these separate is good for maintainability.)

> Add something to explain what the four different boxes represent — four different signing algorithms, or whatever they are.

Yes, I've added an "overview" bit to the description on the right, which hopefully helps with that.

> I noticed that the first and fourth algorithms always produce the same signature, whereas the second and third ones produce different ones each time. Is this a problem, and is it worth calling attention to and saying why this is the case, somehow? E.g. maybe in the first and fourth cases, if you press the sign button more than once, make a message appear saying "signature already generated", and i the other cases show a message saying "generating different signature", or something? I dunno how significant this is...

This is expected: RSA-PSS and ECDSA will generate a different signature value each time, for a given to-be-signed message. The other 2 will always give the same value. I think the message you suggest "signature already generated" would be misleading, since it is generating a new signature each time.

I think trying to explain this in the example is not going to be very helpful, but what I just pushed borrows the trick from the JS interactive examples, where we fade in the new signature value, which hopefully makes the point that a new signature is being generated, even if it's the same as the last one.

Does this help?

Anyway, if you're happy with these updates, please merge this and I'll get on with the examples for encrypt/decrypt.

chrisdavidmills:
I'm happy with these responses Will; all sounds reasonable. I'll merge it.

And thanks @april for all the help. This is awesome.

PR #34, opened 2018-12-15
Add examples for encrypt/decrypt

wbamberg:
@april , @chrisdavidmills - here's a PR adding examples of encrypt/decrypt for the 4 encryption algorithms supported by the Web Crypto API:

* RSA-OAEP
* AES-CTR
* AES-CBC
* AES-GCM

It follows very closely the structure of the previous PR for sign/verify.

In particular I'm interested in knowing whether I've got the params right for the encrypt operations, especially for AES-CTR, which I found a bit obscure. The values I've used come from my reading of [Appendix B of the NIST SP800-38A standard](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38a.pdf#%5B%7B%22num%22%3A70%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22Fit%22%7D%5D), which I've tried to document here: https://developer.mozilla.org/en-US/docs/Web/API/AesCtrParams. But of course it's possible I've made some horrible errors.

### DIFF
--- a/web-crypto/README.md
+++ b/web-crypto/README.md
@@ -3,3 +3,5 @@
 Examples of how to use the [Web Crypto API](https://developer.mozilla.org/docs/Web/API/Web_Crypto_API).
 
 * [sign/verify](sign-verify/index.html): examples showing how to use the [`SubtleCrypto.sign()`](https://developer.mozilla.org/docs/Web/API/SubtleCrypto/sign) and [`SubtleCrypto.verify()`](https://developer.mozilla.org/docs/Web/API/SubtleCrypto/verify) APIs.
+
+* [encrypt/decrypt](encrypt-decrypt/index.html): examples showing how to use the [`SubtleCrypto.encrypt()`](https://developer.mozilla.org/docs/Web/API/SubtleCrypto/encrypt) and [`SubtleCrypto.decrypt()`](https://developer.mozilla.org/docs/Web/API/SubtleCrypto/decrypt) APIs.
--- a/web-crypto/encrypt-decrypt/aes-cbc.js
+++ b/web-crypto/encrypt-decrypt/aes-cbc.js
@@ -23,6 +23,7 @@
   */
   async function encryptMessage(key) {
     let encoded = getMessageEncoding();
+    // The iv must never be reused with a given key.
     iv = window.crypto.getRandomValues(new Uint8Array(16));
     ciphertext = await window.crypto.subtle.encrypt(
       {
--- a/web-crypto/encrypt-decrypt/aes-cbc.js
+++ b/web-crypto/encrypt-decrypt/aes-cbc.js
@@ -33,13 +33,13 @@
       encoded
     );
 
-    let buffer = new Uint8Array(ciphertext, 0, 10);
+    let buffer = new Uint8Array(ciphertext, 0, 5);
     const ciphertextValue = document.querySelector(".aes-cbc .ciphertext-value");
     ciphertextValue.classList.add('fade-in');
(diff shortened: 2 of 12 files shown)


wbamberg:
@chrisdavidmills , I hope I have understood your comments and dealt with them in the latest updates.

chrisdavidmills:
@april, any comments before I merge this?

wbamberg:
@rugk , yeah, I should fix this.

april:
You are using random IVs for AES-GCM, limiting you to 2^32 (or fewer) encrypted messages using that key.

I don't know if we want to call that out, or if it might be better to simply state that the IV should never be reused with a given key.

april:
We may also want to note that neither AES-CBC and AES-CTR have any sort of authentication to them, so you'd to implement some kind of integrity protection alongside them, using HMAC and a different key. It's probably best to nudge people towards using AES-GCM if they're not completely sure of the pitfalls (which are many).

rugk:
> You are using random IVs for AES-GCM and AES-CTR, limiting you to 2^32 (or fewer) encrypted messages using that key.

The question would be, how to solve that? [I've read](https://stackoverflow.com/questions/36760973/why-is-random-iv-fine-for-aes-cbc-but-not-for-aes-gcm) [up on](https://crypto.stackexchange.com/questions/42982/safety-of-random-nonce-with-aes-gcm) that, and [especially ImperialViolet's blog post](https://www.imperialviolet.org/2015/05/16/aeads.html) and it just seems, it is not easy to solve due to the small size of the nonce.

So I guess that topic could be suited for the "security guide" @wbamberg planned to do/(co)write…

rugk:
So maybe write "should not be re-used" and add (as devs automatically think "oh well, let's randomly generate it then") "If generated randomly, there is a chance of X that it is actually getting re-used" where that last part links to a detailed doc about different ways to safely use that nouce?

wbamberg:
I suggest we should avoid adding guidance in the examples. These are examples, not documentation. The documentation is the pages in the Wiki. If we add some normative guidance in here, we then have to duplicate it in the Wiki, and risk things diverging or getting out of date.

Also if we add *some* normative guidance, one danger is that people will think this is a complete guide to using these APIs securely.

This includes things like not reusing IVs and the fact that CTR and CBC don't have authentication. Both these points should be covered in the MDN pages.

So I would prefer to add some boilerplate to these examples pointing people at the MDN pages for more documentation on how to use these APIs, for the benefit of people who have stumbled on the GitHub examples via some other route than the MDN pages.

(Having said that, this is a bit of a grey area, since we do already recommend people should consider 4096-bit RSA keys. But I think definitely things like comparing different algorithms should be out of scope for these examples.)

april:
That's fair, I just worry that once we post examples that people will copy them verbatim without being aware of the problems. I guess we'll have the big warning sign at the top of each page, so hopefully that'll do.  :)

wbamberg:
Hm, then do you think we should also have some boilerplate in these files to try to discourage that?

Something like:

```
/*
The code in this file is intended to illustrate basic usage of these
APIs. It tries to avoid obvious security pitfalls, but should not be
used as-is in your own code.

To use the Web Crypto APIs securely you must have a good understanding
of the security requirements of your system, and the specific security
practices that are needed to meet them.
*/
```

wbamberg:
For the time being I've just added a warning here about not using IVs.

chrisdavidmills:
I think this is OK for now; I'll merge this so we can move forward. We can always add further improvements later on.

PR #35, opened 2018-12-18
Added examples for deriveKey

wbamberg:
@april , @chrisdavidmills - here's a PR adding examples of deriveKey. Currently it just includes examples for ECDH and PBKDF2.

Again it follows the same structure as the others.

I haven't yet added an example for HKDF, because I'm not sure what a good use case would be for it here. In general, AIUI, something that has a single (strongly random) key input, that I can to use to derive multiple keys. But I'm not really sure how to build that into a concrete and fairly simple example. I'd be very happy to hear suggestions...

### DIFF
--- a/web-crypto/derive-key/ecdh.js
+++ b/web-crypto/derive-key/ecdh.js
@@ -98,8 +98,10 @@
     );
   }
 
-  async function setup() {
-    // Generate 2 ECDH key pairs: one for Alice and one for Bob.
+  async function agreeSharedSecretKey() {
+    // Generate 2 ECDH key pairs: one for Alice and one for Bob
+    // In more normal usage, they would generate their key pairs
+    // separately and exchange public keys securely
     let alicesKeyPair = await window.crypto.subtle.generateKey(
       {
         name: "ECDH",
@@ -137,6 +139,6 @@
     });
   }
 
-  setup();
+  agreeSharedSecretKey();
 
 })();
--- a/web-crypto/derive-key/ecdh.js
+++ b/web-crypto/derive-key/ecdh.js
@@ -82,8 +82,8 @@
   - our ECDH private key
   - their ECDH public key
   */
-  async function deriveSecretKey(privateKey, publicKey) {
-    let secretKey = await window.crypto.subtle.deriveKey(
+  function deriveSecretKey(privateKey, publicKey) {
+    return window.crypto.subtle.deriveKey(
       {
         name: "ECDH",
         public: publicKey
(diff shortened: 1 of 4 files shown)


wbamberg:
I've updated the copy in line with the comments.

chrisdavidmills:
@april do you have any comments here?

april:
I'm on PTO (and then holidays) until the 26th. If you want, you can approve it now and I can send you a PR? Alternatively, if you're willing to wait until then I'll take a look at it first thing on the 26th.  :)

chrisdavidmills:
@april I'm totally happy with either. @wbamberg, what would you prefer?

wbamberg:
I'm happy to wait, this isn't blocking me. Thanks for all the reviews, people!

april:
I'm not sure how I feel about the name of the `setup()` function in ecdh.js. I wish it was a bit more explicit on what it does.

I would probably also be a bit more explicit inside `setup()` that although this function generates key pairs for both Alice and Bob, in a more typical operation they would have already securely exchanged public keys and wouldn't know each other's private keys.

How this key exchange happens securely should probably left unexplained, but it can be quite complicated to do.

wbamberg:
> I'm not sure how I feel about the name of the setup() function in ecdh.js. I wish it was a bit more explicit on what it does.

Fair enough. Do you have any suggestions? Perhaps `agreeSecretKey` or `deriveSecretKey`?

> I would probably also be a bit more explicit inside setup() 

Like, adding a bit extra to this comment: https://github.com/mdn/dom-examples/pull/35/files#diff-187496216701805d6bec7d20c43cd1e8R102 ? I agree it's worth doing this. I did have another example that emphasised the separation between Alice and Bob a bit more, but it got complicated, and I decided it wasn't worth the complexity.

april:
> Fair enough. Do you have any suggestions? Perhaps agreeSecretKey or deriveSecretKey?

I think `agreeSecretKey` is fine, or maybe even `agreeSharedSecretKey`.

> Like, adding a bit extra to this comment: https://github.com/mdn/dom-examples/pull/35/files#diff-187496216701805d6bec7d20c43cd1e8R102 ? I agree it's worth doing this. I did have another example that emphasised the separation between Alice and Bob a bit more, but it got complicated, and I decided it wasn't worth the complexity.

Yes, exactly. It doesn't need to be much, maybe:
```javascript
// Generate 2 ECDH key pairs: one for Alice and one for Bob
// In more normal usage, they would generate their key pairs
// separately and exchange public keys securely
```

chrisdavidmills:
Latest changes look good and in accordance to @april 's latest comments. Are we happy to merge this now?

april:
Looks good to me!

chrisdavidmills:
Cool - merging!

PR #36, opened 2018-12-21
Add examples for importKey and exportKey

wbamberg:
Here are some examples for the [`importKey()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/importKey) function.

They're a lot like the others, but this is organised by import format rather than by algorithm.

There's a fair bit of annoying byte-shuffling going on with the PKSC#8 and SPKI ones - it seems like you're supposed to use `TextDecoder` rather than the example I copied from https://developers.google.com/web/updates/2012/06/How-to-convert-ArrayBuffer-to-and-from-String, but I could not get that to work, perhaps because the string does actually contain just binary data and not UTF-8?

### DIFF
--- a/web-crypto/export-key/index.html
+++ b/web-crypto/export-key/index.html
@@ -1,0 +1,51 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Web Crypto API example</title>
+    <link rel="stylesheet" href="style.css">
+  </head>
+
+  <body>
+    <main>
+      <h1>Web Crypto: exportKey</h1>
+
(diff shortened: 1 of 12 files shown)


wbamberg:
I just added examples for `exportKey()` as well - it includes the same formats and algorithms as for `importKey()` but the organization is a bit different, because the interaction is simpler, but the output is bigger :).

wbamberg:
Thanks @chrisdavidmills , I've made the requested change.

wbamberg:
Merging, since this has r+ from Chris. @April , if you see any issues here I will fix them in a follow-up.

PR #38, opened 2019-01-05
Add examples for wrapKey and unwrapKey

wbamberg:
This PR adds some examples for wrapKey and unwrapKey.

Structurally these APIs are a lot like importKey and exportKey, so I've borrowed the same structure from the examples in that PR: https://github.com/mdn/dom-examples/pull/36. But wrapKey/unwrapKey are a bit more complicated because the keys are also encrypted: so wrapKey is like exportKey+encrypt, while unwrapKey is like decrypt+importKey.

For unwrapKey I'm deriving the decryption key from a password, like the PBKDF2 example in the deriveKey PR (https://github.com/mdn/dom-examples/pull/35). I only used a couple of variants for unwrapKey, so as to keep things manageable, but the 2 examples cover (I hope) most of the points of variation in this API.

If you're not familiar with these APIs, it might be easier to review this PR after I have revised the MDN pages for them?

### DIFF
--- a/web-crypto/unwrap-key/index.html
+++ b/web-crypto/unwrap-key/index.html
@@ -1,0 +1,60 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Web Crypto API example</title>
+    <link rel="stylesheet" href="style.css">
+  </head>
+
+  <body>
+    <main>
+      <h1>Web Crypto: unwrapKey</h1>
+
+      <section class="description">
+        <p>This page shows the use of the <code>unwrapKey()</code> function of the <a href="https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API">Web Crypto API</a>. It contains two separate examples:</p>
+        <ul>
+          <li>Raw/AES-GCM/AES-KW: unwrap an AES-GCM encryption key, that was encrypted using AES-KW and exported in "raw" format.</li>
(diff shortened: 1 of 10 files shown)


wbamberg:
Good suggestions @chrisdavidmills , I've had a go at those updates.

april:
I'll try to review these this week, but I'm out on Friday and a bit swamped for the next couple days.

I'll try to aim for sometime tomorrow morning. Sorry for the delay!

chrisdavidmills:
@april no worries at all if the reviews come slightly later; we really appreciate your time on this.

april:
I like the idea of the dialog also stating what the password is, such as:

```javascript
const password = window.prompt("Enter your password. The working password is “correct horse battery staple”.");
```

The code (such as raw.js) also has a lot of mixing of styles, using both `function` and arrow syntax, particularly in the event handlers. I forget if that had been okay or not.  :)

Otherwise, I think things look really great. Nice job, and sorry for taking so long on this.

wbamberg:
Thanks April!

> I like the idea of the dialog also stating what the password is

I've made those changes in unwrap-key.

> The code (such as raw.js) also has a lot of mixing of styles, using both function and arrow syntax

As far as I know I'm using arrow syntax for anonymous functions but function syntax for named functions, and this seems to be pretty common in the examples we have in MDN, but I'm happy to go with whatever our style guide says. Having said that this is also the style of the Web Crypto examples already merged, so we'd have to go back and fix those too.

PR #39, opened 2019-01-08
Add examples for deriveBits

wbamberg:
Examples for the [`deriveBits()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/deriveBits) API. This is like [`deriveKey()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/deriveKey), except that `deriveKey()` also imports the bits as a `CryptoKey`, enabling you to actually use the derived key to do things like encryption.

So these examples are based on the [examples for `deriveKey()`](https://github.com/mdn/dom-examples/tree/master/web-crypto/derive-key), except they're simpler because you can't actually use the derived bits to encrypt.

I think this is the last example we need :).

### DIFF
--- a/web-crypto/derive-bits/ecdh.js
+++ b/web-crypto/derive-bits/ecdh.js
@@ -1,0 +1,56 @@
+(() => {
+
+  async function deriveSharedSecret(privateKey, publicKey) {
+    const sharedSecret = await window.crypto.subtle.deriveBits(
+      {
+        name: "ECDH",
+        namedCurve: "P-384",
+        public: publicKey
+      },
+      privateKey,
+      128
+    );
+
+    const buffer = new Uint8Array(sharedSecret, 0, 5);
+    const sharedSecretValue = document.querySelector(".ecdh .derived-bits-value");
+    sharedSecretValue.classList.add("fade-in");
+    sharedSecretValue.addEventListener("animationend", () => {
+      sharedSecretValue.classList.remove("fade-in");
+    });
+    sharedSecretValue.textContent = `${buffer}...[${sharedSecret.byteLength} bytes total]`;
+  }
+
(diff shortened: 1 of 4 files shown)


chrisdavidmills:
Once @april has given this a review and approved, I think we can merge it.

wbamberg:
>  maybe you should link the mention of "deriveBits()" to the appropriate reference page on MDN, when it is available

It's a good suggestion and I don't know why I didn't do this (and BTW deriveBits is available now, I wrote the page yesterday: https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/deriveBits. But it would apply to all the examples (e.g. https://github.com/mdn/dom-examples/blob/master/web-crypto/derive-key/index.html#L14), so it would be much easier to do this as a single follow-up PR, rather than amend every open PR with this change, so can we do that?

april:
Overall it looks fine to me, other than to say that if we're deriving a 256-bit key in `deriveKey()` then we should probably do the same in `deriveBits()`. Currently ecdh.js is deriving a 128-bit key.

wbamberg:
Thanks again for the reviews!

> Currently ecdh.js is deriving a 128-bit key.

Again don't know I did this :). I've updated the PR.

>> it would be much easier to do this as a single follow-up PR, rather than amend every open PR with this change, so can we do that?

> Yup, absolutely.

I've filed https://github.com/mdn/dom-examples/issues/40 so we don't forget.

chrisdavidmills:
Cool, all looks good then. Merging.

web-crypto/derive-bits/ecdh.js
```diff
@@ -8,7 +8,7 @@
         public: publicKey
```
nowakwp, line 8: mdn:master

PR #46, opened 2019-05-24
Remove unused variable declarations in encrypt-decrypt example

eliranlevi:
Removing these unused variables might improve readability and reduce confusion :)


chrisdavidmills:
cc @wbamberg

wbamberg:
@eliranlevi , thank you!

This is definitely a good change, but I think it would also be good to update the comments that currently say `Fetch the encoded message and decrypt it.` to something like `Fetch the ciphertext and decrypt it.`.

eliranlevi:
@wbamberg thanks for the feedback! I've updated the PR accordingly. Please let me know if it looks better now or if further changes are required.

I'm glad I could contribute :)

PR #48, opened 2019-06-25
Remove " in object's parameter

siprikorea:


chrisdavidmills:
@wbamberg can you review this please? Thanks in advance.

wbamberg:
Thanks for the contribution @siprikorea ! As far as I know the existing code is correct. The change you've made here is fine, and probably better, but I think if we want to do this we should do it consistently at least across the examples under web-crypto. There are lots of other places where properties are quoted (e.g. https://github.com/mdn/dom-examples/blob/master/web-crypto/unwrap-key/pkcs8.js#L128-L131) and it will be weird to be inconsistent.

We should also update all the Wiki pages that copy bits of this code, for example https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/importKey#Raw_import.

wbamberg:
Closing this as it seems to be inactive.

PR #29, opened 2018-11-06
Add chroma keying

wbamberg:
This migrates the live example from developer.mozilla.org/samples to GitHub, in an attempt to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1501297, specifically as Julien suggests in comment 3: https://bugzilla.mozilla.org/show_bug.cgi?id=1501297#c3..

It adds "canvas/chroma-keying", that contains an HTML file and a JS file corresponding to the files discussed in https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Manipulating_video_using_canvas, as well as the media files used in the example.

I've tested this at https://wbamberg.github.io/dom-examples/canvas/chroma-keying/ and it seems to work fine.

PR #51, opened 2019-11-19
Avoid Security Error

severo:
See https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image.

### DIFF
--- a/canvas/chroma-keying/index.html
+++ b/canvas/chroma-keying/index.html
@@ -22,7 +22,7 @@
 
   <body>
     <div>
-      <video id="video" src="media/video.mp4" controls="true"/>
+      <video id="video" src="media/video.mp4" controls="true" crossorigin="anonymous"/>
     </div>
     <div>
       <canvas id="c1" width="160" height="96"></canvas>


wbamberg:
Thanks for the PR!

Can you please elaborate what the security error is here? The page this example is used in (https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Manipulating_video_using_canvas) works fine for me in Firefox and Chrome, with no relevant console errors.

severo:
Sure, sorry. The error only appears when the video src is located on another domain than the code.
The error is:

> SecurityError: The operation is insecure.

And it comes from calling [`getImageData`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/getImageData), as stated in the [Exceptions](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/getImageData#Exceptions) section.

See for example https://observablehq.com/@severo/manipulating-video-using-canvas-mdn-tutorial, where the code is on observablehq.com, whereas the video comes from raw.githubusercontent.com. I had to add the `crossorigin="anonymous"` attribute to the `<video>` element. Same exception on https://observablehq.com/@severo/mdn-video-and-canvas-tutorial if you remove `crossorigin="anonymous"`.

wbamberg:
Thanks for the clarification!

This example exists really for the sake of the MDN page at https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Manipulating_video_using_canvas, and it works fine there.

As I understand https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image, making the example work cross-origin is not as simple as just adding `anonymous`, you also have to configure the server appropriately, so it's not a complete solution in itself.

In fact, the reason we added the example to GitHub was exactly to fix this issue (https://github.com/mdn/dom-examples/pull/29, https://bugzilla.mozilla.org/show_bug.cgi?id=1501297) without having to change the server.

So I think this PR is out of scope for us.

severo:
OK, I understand, you're right. Thanks for taking time to review and comment.

PR #57, opened 2020-05-09
simulate multi-touch

zjffun:
PC browsers can't see all effect of this demo, so I added an emulator.


zjffun:
Doesn't seem to help.

PR #58, opened 2020-05-12
Refactor AbortController example

ehpc:
Promise hell anti-pattern refactored.
Vars replaced with const/let.
Inline styles replaced with classes.


chrisdavidmills:
@ehpc I've checked this out, and it seems to work really nicely. Thanks for the contribution.

PR #67, opened 2020-10-27
Create Canvas pixel manipulation examples

gsouquet:
This work is related to mdn/sprints#1127

Moving the three examples from https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Pixel_manipulation_with_canvas to this repository to avoid CORS issue when fetching the image


chrisdavidmills (changes requested):
Hi @gsouquet !

More great work here. A few review comments, but nothing major:

1. On the color manipulation example:

* The invert buton can be clicked multiple times to toggle the invert on and off, but when you click grayscale it goes to gray, and subsequent clicks don't toggle it off and on again. Would it make sense to "fix" this, or explain why not somewhere?
* UK spelling in the title - "colour", update to "color"
* 2nd half of the title has every word capitalized; MDN style is to use sentence case, e.g. "Canvas pixel manipulation". Same in the other example titles.

2. Color picker example:

* This is really cool; the only worry I had is that when you click a coloured pixel, the colour box to the right doesn't seem to stick with that color reliably. Maybe it'd be more obvious what is going on if you had two boxes, with labels of "Currently hovered color" and "Selected color", the first for the currently hovered color and the second for the color that is copied?
* Maybe you could also copy the clicked color to the user's clipboard? ;-)
 
3. The image smoothing example

* It is hard to see what the effect is of cheking and unchecking the box, both in Firefox and Chrome. Any way this can be improved on? Maybe with a differen image?
* In Firefox you get this warning — "Use of mozImageSmoothingEnabled is deprecated. Please use the unprefixed imageSmoothingEnabled property instead."

gsouquet:
Thank you for your feedback @chrisdavidmills !
Addressed mostly everything but I left the Clipboard API out of it but created some beautiful tables instead ahah

chrisdavidmills (changes requested):
@gsouquet much better, thanks!

Just a couple of nitpicks:

1. In the color manipulation example, can you update it so that the "active button" is highlighted in some way? Make it more obvious that it works like a set of radio buttons rather than individual toggle switches.
2. In the color picker example `<title>`, update "Picker" to "picker".

chrisdavidmills (approved):
Yes, looks good to me!

PR #68, opened 2020-10-27
Add Fullscreen API example

tarunbatra:
The PR adds an example for [Full screen API](https://wiki.developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API) since [the existing one](https://media.prod.mdn.mozit.cloud/samples/domref/fullscreen.html) is broken.

See https://github.com/mdn/sprints/issues/3828

__Thoughts:__ I have tried to keep the example similar to the example code shown in the API doc. So the example listens to `ENTER` key and nothing more. Can add frills like buttons and event log if that enhances the example's utility.

Another thing to consider is would the key binding work when used with`{{EmbedGHLiveSample()}}` macro?

### DIFF
--- a/fullscreen-api/README.md
+++ b/fullscreen-api/README.md
@@ -1,6 +1,6 @@
 
 # Fullscreen API example
 
-The example toggles the video in and out of fullscreen mode on the press of ENTER key.
+This example allows you to toggle the video in and out of fullscreen mode by pressing the <kbd>Enter</kbd> key.
 
 [See the example live](https://mdn.github.io/dom-examples/fullscreen-api/).
--- a/fullscreen-api/README.md
+++ b/fullscreen-api/README.md
@@ -1,0 +1,6 @@
+
+# Fullscreen API example
+
+The example toggles the video in and out of fullscreen mode on the press of ENTER key.
+
+[See the example live](https://mdn.github.io/dom-examples/fullscreen-api/).
--- a/fullscreen-api/index.html
+++ b/fullscreen-api/index.html
@@ -9,9 +9,9 @@
 
 <body>
   <div id="contents">
-    <p>A quick example for the use of the <code>Fullscreen</code> API. Press <code>ENTER</code> key to toggle Full screen.</p>
+    <p>A quick example to demonstrate usage of the <a href="https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API">Fullscreen API</a>. Press <kbd>Enter</kbd> to toggle fullscreen mode.</p>
 
-    <video src="assets/bigbuckbunny.mp4" id="video" muted autoplay width="300"></video>
+    <video src="assets/bigbuckbunny.mp4" id="video" controls width="480"></video>
 
     <div id="credits"><a href="https://peach.blender.org/download/">Video by Blender</a>;
       <a href="https://peach.blender.org/about/">licensed CC-BY 3.0</a></div>
--- a/fullscreen-api/index.html
+++ b/fullscreen-api/index.html
@@ -1,0 +1,21 @@
+<!doctype html>
+<html>
+
+<head>
+  <title>Fullscreen API Example</title>
+  <script src="main.js"></script>
+  <link href="main.css" rel="stylesheet">
+</head>
+
+<body>
+  <div id="contents">
+    <p>A quick example for the use of the <code>Fullscreen</code> API. Press <code>ENTER</code> key to toggle Full screen.</p>
+
+    <video src="assets/bigbuckbunny.mp4" id="video" muted autoplay width="300"></video>
+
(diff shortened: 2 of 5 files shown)


fullscreen-api/README.md
```diff
@@ -0,0 +1,6 @@
+
+# Fullscreen API example
+
+The example toggles the video in and out of fullscreen mode on the press of ENTER key.
```
chrisdavidmills, line 4: Rephrase to "This example allows you to toggle the video in and out of fullscreen mode by pessing the <kbd>Enter</kbd> key."

fullscreen-api/index.html
```diff
@@ -0,0 +1,21 @@
...
+
+<head>
+  <title>Fullscreen API Example</title>
+  <script src="main.js"></script>
+  <link href="main.css" rel="stylesheet">
+</head>
+
+<body>
+  <div id="contents">
+    <p>A quick example for the use of the <code>Fullscreen</code> API. Press <code>ENTER</code> key to toggle Full screen.</p>
```
chrisdavidmills, line 12: Rephrase to "A quick example to demonstrate usage of the <a href="https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API">Fullscreen API</a> API. Press <kbd>Enter</kbd> to toggle fullscreen mode."

fullscreen-api/main.css
```diff
@@ -0,0 +1,9 @@
+#contents {
+	width: 600px;
```
chrisdavidmills, line 2: Declaration not really needed?

fullscreen-api/index.html
```diff
@@ -0,0 +1,21 @@
...
+
+    <video src="assets/bigbuckbunny.mp4" id="video" muted autoplay width="300"></video>
```
chrisdavidmills, line 14: use a more obvious width, like 320 or 480? Also provide the `controls` attribute? Also remove the autoplay attribute — this is generally seen as a bad practice.

chrisdavidmills: Also, remove `muted`?

fullscreen-api/main.css
```diff
@@ -0,0 +1,9 @@
...
+	font: 14px "Open Sans", sans-serif;
+}
+
```
chrisdavidmills, line 5: Add a rule to size the video as `max-width: 100%` to make it more responsive on small screens?

fullscreen-api/main.js
```diff
@@ -0,0 +1,28 @@
+window.addEventListener("load", startup, false);
+
+function startup() {
+  // Get the reference to video
+  const video = document.getElementById("video");
+
+  // On pressing ENTER call toggleFullScreen method
+  document.addEventListener("keypress", function(e) {
+    if (e.keyCode === 13) {
```
chrisdavidmills, line 9: AFAIK, `keyCode` is deprecated, and `key` is more the best practice these days. Update it to use `key` instead?

chrisdavidmills (changes requested):
Hi @tarunbatra ! I've just got a few nitpicks for you to address, but nothing too major. Thanks again for taking the time to do this.

fullscreen-api/main.css
tarunbatra: Yeah, I used Picture in picture example as a base and left a lot of things like these untouched. Updated now.

fullscreen-api/main.css
tarunbatra: I added the rule, but it doesn't seem to work when I tested. And I'm not really a css guy. 😅

fullscreen-api/main.js
tarunbatra: Yes this makes sense. I was trying to stay close to the example in MDN docs which used `keyCode`. I'll go ahead and update it as well I think.

tarunbatra:
Hey @chrisdavidmills thanks for the review. I incorporated the changes suggested and  updated the PR.
I still have one doubt though, as mentioned in the description of the PR:
 > ...would the key binding work when used with{{EmbedGHLiveSample()}} macro?

Can you help me with understanding that part too? Thanks.

chrisdavidmills:
> Hey @chrisdavidmills thanks for the review. I incorporated the changes suggested and updated the PR.

This is great, thanks @tarunbatra ! The changes all look fine. The video CSS that you are not so sure about also works fine — when you make your browser window really narrow, the video shrinks so that it stays neatly inside the browser window. This is the behavior I was intending.

> I still have one doubt though, as mentioned in the description of the PR:
> 
> > ...would the key binding work when used with{{EmbedGHLiveSample()}} macro?
> 
> Can you help me with understanding that part too? Thanks.

I _think_ this should still work, but it might not work unless the user is focused inside the iframe. Shall we try it, and if it is a problem, we can instead add an external button to activate fullscreen in addition to the Enter key?

PR #75, opened 2020-11-29
Fix image smoothing properties in canvas zoom demo

nickgaya:
This example demonstrates the use of canvas elements to zoom in on a part of an image.  This code is used on MDN [here](https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Pixel_manipulation_with_canvas#Zooming_and_anti-aliasing).

This PR fixes a bug in the code whereby the "smoothed" canvas actually has image smoothing disabled, while the "pixelated" canvas has image smoothing enabled by default.


chrisdavidmills:
LGTM, thanks for the fix!

PR #71, opened 2020-11-03
corrected the typo

SobhanBera:
I have corrected the desired typo you said. Now you can merge if this is correct.
Thanks

PR #78, opened 2021-01-10
Revert passing signatureValue and fix formatting

zu3st:
This fixes the regression introduced in #71 and reintroduces the old formatting.


chrisdavidmills (approved):
Thanks for catching this.

PR #84, opened 2021-11-11
Fix resize observer examples

wbamberg:
In Firefox, https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry/contentBoxSize used to be a single entry, while the standard (and Chrome) implemented it as an array of entries.

See also https://github.com/mdn/content/pull/3180.

Firefox recently fixed this: https://bugzilla.mozilla.org/show_bug.cgi?id=1689645 .

In dom-examples, one of the examples only worked with the old-Firefox implementation, and the other incorrectly claimed that Chrome's version was non-standard.

This PR is supposed to fix both those issues.

### DIFF
--- a/resize-observer/resize-observer-border-radius.html
+++ b/resize-observer/resize-observer-border-radius.html
@@ -28,9 +28,16 @@
     <script>
       const resizeObserver = new ResizeObserver(entries => {
         for (let entry of entries) {
-          if(entry.contentBoxSize) {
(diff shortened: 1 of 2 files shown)


wbamberg:
@Rumyra , could I get a review of this please? Thanks!

Rumyra:
Hi @wbamberg apologies for missing this - all looks good to me 👍

Rumyra (approved):

PR #86, opened 2021-11-30
Add onresize event handler example

AnilSeervi:
### Changes
- Adds `onresize` event handler example in "resize-event" directory
- Update README with description of the example
- update `http` links in README to `https`

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,19 +5,19 @@ Code examples that accompany various MDN DOM and Web API documentation pages.
 
 * The "auxclick" directory contains a simple example demonstrating the new <code>auxclick</code> event type. See [GlobalEventHandlers.auxclick](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onauxclick) for more details, or [run the example live](https://mdn.github.io/dom-examples/auxclick/).
 
-* The "canvas" directory contains an example "chroma-keying" demonstrating how to use the Canvas API to manipulate videos: see [Manipulating video using canvas](https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Manipulating_video_using_canvas) or [run the example live](http://mdn.github.io/dom-examples/canvas/chroma-keying/).
+* The "canvas" directory contains an example "chroma-keying" demonstrating how to use the Canvas API to manipulate videos: see [Manipulating video using canvas](https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Manipulating_video_using_canvas) or [run the example live](https://mdn.github.io/dom-examples/canvas/chroma-keying/).
 
-* The "channel-messaging-basic" directory contains a simple example demonstrating the basics of channel messaging; see [Channel Messaging API](https://developer.mozilla.org/en-US/docs/Web/API/Channel_Messaging_API) or [run the example live](http://mdn.github.io/dom-examples/channel-messaging-basic/).
+* The "channel-messaging-basic" directory contains a simple example demonstrating the basics of channel messaging; see [Channel Messaging API](https://developer.mozilla.org/en-US/docs/Web/API/Channel_Messaging_API) or [run the example live](https://mdn.github.io/dom-examples/channel-messaging-basic/).
 
-* The "channel-messaging-multimessage" directory contains another channel messaging demo, showing how multiple messages can be sent between browsing contexts. See [Channel Messaging API](https://developer.mozilla.org/en-US/docs/Web/API/Channel_Messaging_API) for more details. [Run the demo live](http://mdn.github.io/dom-examples/channel-messaging-multimessage/).
+* The "channel-messaging-multimessage" directory contains another channel messaging demo, showing how multiple messages can be sent between browsing contexts. See [Channel Messaging API](https://developer.mozilla.org/en-US/docs/Web/API/Channel_Messaging_API) for more details. [Run the demo live](https://mdn.github.io/dom-examples/channel-messaging-multimessage/).
 
 * The "drag-and-drop" directory is for examples and demos of the [HTML Drag and Drop](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API) standard.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -29,13 +29,15 @@ Code examples that accompany various MDN DOM and Web API documentation pages.
...
+* The "reporting-api" directory contains a couple of simple demos to show usage of the Reprting API. You can find more explanation of how the API works in the main MDN [Reporting API](https://developer.mozilla.org/en-US/docs/Web/API/Reporting_API) docs. [Run the deprecation report demo live](https://mdn.github.io/dom-examples/reporting-api/deprecation_report.html).
 
-* The "screenleft-screentop" directory contains a demo to show how you could use the [Window.screenLeft](https://developer.mozilla.org/en-US/docs/Web/API/Window/screenLeft) and [Window.screenTop](https://developer.mozilla.org/en-US/docs/Web/API/Window/screenTop) properties to draw a circle on a canvas that always stays in the same physical place on the screen when you move your browser window. [Run the demo live](http://mdn.github.io/dom-examples/screenleft-screentop/).
+* The "resize-event" directory contains a simple demo to show how you can use the [onresize](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onresize) event handler. Resize the browser window either by height or width to see the size of your current window. [Run the demo live](https://mdn.github.io/dom-examples/resize-event).
```
wbamberg, line 36: We're currently at the start of a project where we will merge event pages, so instead of `onXYZ` and `XYZ_event`, we will just have `XYZ_event`. So it will be more future-proof for this to talk about "the `resize` event and link to https://developer.mozilla.org/en-US/docs/Web/API/Window/resize_event, rather than talk about `onresize`.

-> https://github.com/mdn/content/discussions/9098

resize-event/index.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta name="description" content="A resize event demo on MDN" />
```
wbamberg, line 7: Not sure which of these we need. From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta it looks like we don't need `charset` or `http-equiv` anyway.

resize-event/index.html
```diff
@@ -0,0 +1,27 @@
...
+    <script>
+      const heightOutput = document.querySelector("#height");
+      const widthOutput = document.querySelector("#width");
+
+      function resize() {
+        heightOutput.textContent = window.innerHeight;
+        widthOutput.textContent = window.innerWidth;
+      }
+
+      window.onresize = resize;
```
wbamberg, line 24: I'd prefer us to use `addEventListener` here as we tend to recommend that method (https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events#what_mechanism_should_i_use, https://developer.mozilla.org/en-US/docs/Web/API/eventtarget/addeventlistener).

wbamberg (changes requested):
Thank you @AnilSeervi ! I had a few comments but this in general looks great.

I see though that you're using your `master` branch for this PR - maybe you want to use a different branch, in case you ever want to make more PRs to this repo?

wbamberg:
This is part of the fix for https://github.com/mdn/content/issues/10683.

resize-event/index.html
wbamberg: actually maybe that doesn't make sense since this is for the `onresize` property!

AnilSeervi: Yes, but there's the same example on https://developer.mozilla.org/en-US/docs/Web/API/Window/resize_event page with a description :

> The following example reports the window size each time it is resized. Bear in mind that since the example is running in an <iframe>, you'll need to actually get the <iframe> to resize before you see an effect.

And it also shows how to add event listener using `addEventListener` - https://developer.mozilla.org/en-US/docs/Web/API/Window/resize_event#addeventlistener_equivalent

AnilSeervi:
There is actually a macro for what I suggested in [comment#10683](https://github.com/mdn/content/issues/10683#issuecomment-977384432).
I found it here - https://github.com/mdn/content/blob/main/files/en-us/web/api/indexeddb_api/using_indexeddb/index.md?plain=1#L1348
But the link it generated seems to be broken. - mdn/content#10941

resize-event/index.html
wbamberg: Hey @AnilSeervi , and sorry to be so slow to look at this. I think this example should use `addEventListener`.

Quite soon we will merge the `onresize` and `resize_event` pages together, and then we will document `addEventListener` as the preferred form.

For examples embedded in the page, it makes sense to use the correct form for the page - so I mean, the `onresize` page should use `onresize`, and we'll take care of any updates when merging the pages. But because this is going to be a link from `onresize`, it'll be a pain to have to update it over here when we change the pages in mdn/content. So it's better if we make this one use `addEventListener` from the start.

Does that make sense?

wbamberg:
> There is actually a macro for what I suggested in [comment#10683](https://github.com/mdn/content/issues/10683#issuecomment-977384432). I found it here - https://github.com/mdn/content/blob/main/files/en-us/web/api/indexeddb_api/using_indexeddb/index.md?plain=1#L1348 But the link it generated seems to be broken. - [mdn/content#10941](https://github.com/mdn/content/issues/10941)

Interesting!

It looks like this is broken in the IndexedDB page because you must have previously called `EmbedLiveSample` - that tells Yari to generate the sample page. Otherwise the sample page doesn't exist. You can see it does work in https://developer.mozilla.org/en-US/docs/Web/API/datatransferitemlist/clear#example.

wbamberg (approved):
👍 thank you @AnilSeervi , this looks good to me. Unfortunately I introduced a merge conflict, let me know if you want me to try to fix it (or fix it yourself if you prefer).

AnilSeervi (commented):
I've resolved the conflict @wbamberg.
You can now go ahead and merge this :)

wbamberg:
Thanks @AnilSeervi !

PR #87, opened 2021-12-22
clearing fade-in animation eventListeners

ordinaryperson420:
There is an `EventListeners` that keep being added every time when `fade-in` animation plays. Their only purpose is to remove a `className` that triggers element animation via css. And they were never removed. I considered adding `{ once: true }` to prevent this.


bsmth (approved):
Super, thank you!

PR #89, opened 2022-01-03
Fix pick coordinates relative to the image

nsauzede:
Current code uses absolute mouse coordinates instead of relative to the canvas, which in turn makes the actual hovered/selected color inaccurate (there's an x/y offset that reads pixel further towards top/left direction)
This diff hasn't been tested in the exact tutorial context, but seems to fix the bug and work fine in a simpler, similar unit test (which had the exact same issue)

### DIFF
--- a/canvas/pixel-manipulation/color-picker.js
+++ b/canvas/pixel-manipulation/color-picker.js
@@ -12,8 +12,9 @@ var selectedColor = document.getElementById('selected-color');
 
 
 function pick(event, destination) {
-  var x = event.layerX;
-  var y = event.layerY;
+  var rect = event.target.getBoundingClientRect();
+  var x = event.layerX - rect.left;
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):

sideshowbarker:
@nsauzede Sorry for the very long delay in anybody getting around to reviewing this. I’ve approved it — but because there’s been another change to the sources in the meantime, there’s now a merge conflict that needs to be resolved.

If you can make time to take a look and to resolve the merge conflict, then I can merge this once the conflict is resolved.

nsauzede:
Hi, thanks for the approval @sideshowbarker 
However, looking at the conflicts, it seems like upstream already fixed my bug with the same spirit, just with slightly different syntax
I guess it's safe for me to close this PR now as "obsolete"

PR #90, opened 2022-01-15
Add indexeddb demo

wbamberg:
To fix https://github.com/mdn/content/issues/10941, this PR transfers the demo currently at https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Using_IndexedDB#full_indexeddb_example, because I don't think long-form demos work very well as huge code blocks in the page.

The code needs some attention, but I thought it would be better to transfer it first, then improve it once it is here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ Code examples that accompany various MDN DOM and Web API documentation pages.
 
 * The "fullscreen-api" directory is for examples and demos of the [Fullscreen API](https://wiki.developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API). Run the [example live](http://mdn.github.io/dom-examples/fullscreen-api/).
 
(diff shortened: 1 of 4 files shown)


Elchi3 (approved):
Good idea to move this large chunk of code here but haven't reviewed it itself. As you say, improvements/fixes can come after the initial import.

Elchi3 (changes requested):
Oh, approved too fast! There is a commit to the resize observer example which shouldn't be part of this PR

wbamberg:
Oops, thank you for catching that @Elchi3 . Should be fixed now?

wbamberg:
> Good idea to move this large chunk of code here but haven't reviewed it itself. As you say, improvements/fixes can come after the initial import.

I have checked it works, but it is very much in need of improvement. But I thought it would be much easier to review by moving it as-is, then applying changes in a separate PR.

Elchi3 (approved):
Yup, all good now

PR #92, opened 2022-02-02
Fix stream simple-pump example.

BMorearty:
The simple-pump example wasn't convincing because when you
view source, it showed that both images' src attribute
were tortoise.jpg. By inspection, you couldn't tell that
the stream was modifying the second image's src.

I fixed this by changing the second image's src to ".".

Tested by running it in Firefox and seeing that the second
image still shows a tortoise.


sideshowbarker (approved):

sideshowbarker:
Brian, thanks much, and congrats on landing your first docs change here — welcome aboard 🎉
 (and sorry the long delay in getting this one merged)

PR #97, opened 2022-03-27
Streams/simple-writer: fixed block character in the output.

OnkarRuikar:
There are unwanted block characters in the output:
screenshot

With `utf-16` encoding for the decoder, it fixes the issue:
```
const decoder = new TextDecoder("utf-16");
```


sideshowbarker:
See my comment at https://github.com/mdn/content/pull/14371#issuecomment-1080075957

sideshowbarker (approved):

Issue #99, opened 2022-04-18
pixel-manipulation in bottom area


Bechma:
I was reading the [tutorial](https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Pixel_manipulation_with_canvas) assigned to [pixel-manipulation/color-picker](https://github.com/mdn/dom-examples/tree/master/canvas/pixel-manipulation) and it's not working as expected in Edge(100.0.1185.44) and Chrome(100.0.4896.88)

![image](https://user-images.githubusercontent.com/19294519/163833624-fe3c1fc3-33f8-4036-aa92-b0072a75d0ed.png)

I found out that when you hover the mouse in the far top of the canvas, it shows y=33, but when you hover down to the bottom part, it shows y=259. This situation makes that the bottom part of the image can't be picked.

PR #100, opened 2022-04-18
fix wrong x and y in color-picker

Bechma:
Now the far left pixel will be 0 and not 11. Also, the far top pixel will be 0 instead of 33.

Related https://github.com/mdn/dom-examples/issues/99
(Tested in chrome)


wbamberg (approved):
👍 thank you @Bechma ! And sorry to take so long to review this one.

PR #101, opened 2022-05-13
moving simple-web-worker into dom examples

Rumyra:
moving simple-web-worker example into dom examples


.gitignore
```diff
@@ -0,0 +1 @@
+.DS_Store
```
schalkneethling, line 1:
```suggestion
.DS_Store

```

schalkneethling (approved):
🎉

PR #102, opened 2022-05-19
Fix broken picture-in-picture example credit links

ncnlinh:
### Why

https://developer.mozilla.org/en-US/docs/Web/API/Picture-in-Picture_API

Clicking on the links here will break the iframe. Adding `target="_blank"` allows them to open in a new  tab of the parent frame, which is more desirable.


wbamberg (approved):
👍 thank you @ncnlinh !

PR #106, opened 2022-05-27
streams/simple-writer: remove invalid parameter

sxyazi:
There is a `{ stream: true }` does not exists on the `TextEncoder.encode` parameter list.


lucalves (approved):

teoli2003 (approved):
You are right! Thanks!

PR #125, opened 2022-07-07
Fix incorrect information

MrBrain295:
https://developer.chrome.com/blog/auxclick/


teoli2003:
I've fixed it to make it timeless. We don't have a reliable process to update these readme files, so let's make them timeless (or time-agnostic).

PR #129, opened 2022-07-07
Partially modernize web storage example

MrBrain295:


teoli2003 (approved):
Please update js files first. These changes don't bring much. Ok, as part of a larger cleaning, but creating (and reviewing) a whole PR only for such a change is a bit overkill

PR #141, opened 2022-07-29
Fix a Nu HTML checker warning

MrBrain295:


teoli2003:
Can you elaborate on what this PR changes? This is not an HTML change and I don't see what it does to HTML validation warning.

MrBrain295:
The font size was set to be reliant on the viewport but this changes the way the font size is defined.

teoli2003:
Yes, I see this. I still don't understand why this has an impact on HTML validation.

MrBrain295:
Sorry, last time I ran it I remember seeing an error but they doesn’t seem to be the case.

Issue #155, opened 2022-08-17
Improper Distance Calculation in Zoom Event


InkboxSoftware:
https://github.com/mdn/dom-examples/blob/master/pointerevents/Pinch_zoom_gestures.html

Line 79 of this script should be:
`var curDiff = Math.sqrt(Math.pow(evCache[1].clientX - evCache[0].clientX, 2) + Math.pow(evCache[1].clientY - evCache[0].clientY, 2));`

Currently line 79 is only calculating the difference between the X coordinates of the touches, but it should actually be measuring the distance between the two points. In the case of the difference of the X coordinates not changing or even decreasing, but the overall distance between the two touches is increasing then the current line 79 would not register a zoom out, but a zoom in. The line written above prevents such confusion.

teoli2003:
Yep, Euclidian distance. Do you want to create a PR?

InkboxSoftware:
I've just submitted a pull request. I'm not quite so familiar with GitHub, so I hope I did it correctly.

caugner:
Resolved via https://github.com/mdn/dom-examples/issues/155.

benjamingwynn:
Just ran into this. Does this not relate to https://github.com/mdn/dom-examples/issues/9 ?

PR #156, opened 2022-08-18
Update Distance Calculation in Zoom Event

InkboxSoftware:
Updated check between distance of two touches in line 79 to calculate the Euclidean distance, instead of what was before only checking the distance between the X coordinate of the two touches.


teoli2003 (approved):
All fine! Thank you!

teoli2003:
Thanks a lot!

Looks good to me.

Congratulations upon your first merged commit here! Welcome aboard! :tada:

caugner:
Fixed https://github.com/mdn/dom-examples/issues/155.

PR #161, opened 2022-09-10
Use the closest method instead of parentNode.parentNode

dezman:
This way if someone (like me) edits the file, the code will still remove the correct clip recording (in most cases).


teoli2003 (approved):
Yes, `closest()` didn't exist when this example was written. I removed the intermediate variable (that should have been `const`-declared anyway.

There is much more cleaning that could be done on this file, but this is for another time.

teoli2003:
Thanks a lot!

Looks good to me.

Congratulations upon your first merged commit here! Welcome aboard! :tada:

dezman:
Awesome!

PR #172, opened 2022-09-20
simplify / correct markup of dialog example

scottaohara:
- adds `lang=en` to the `html` element
- removes unnecessary use of `section` element from `dialog`
- swap invalid use of `menu` element with `div`


htmldialogelement-basic/index.html
```diff
@@ -9,25 +9,25 @@
   <!-- Simple pop-up dialog box, containing a form -->
   <dialog id="favDialog">
     <form method="dialog">
-      <section>
-        <p><label for="favAnimal">Favorite animal:</label>
+      <p>
```
juanpprieto, line 12: Perhaps a `<fieldset>` tag is better suited here?

scottaohara: that wouldn't be useful in this example.

wbamberg (approved):
👍 thank you @scottaohara and sorry to be so slow to merge this. We are not great at monitoring this repo.

PR #180, opened 2022-10-14
fix(importPublicKey): extract pemContents

sixcolors:
Import a PEM encoded RSA private key function importPrivateKey(pem) did not properly extract the pemContents.

Corrected length -1 and added trim() to remove newline/whitespace if present.

related https://github.com/mdn/content/pull/21540


sixcolors:
Invalid. Sorry about that.

PR #188, opened 2022-11-28
Fix Nu HTML checker errors on  channel-messaging-basic

MrBrain295:
And run prettier.


teoli2003 (approved):
There is more0 work to be done here (avoiding var for example) but this is already better. Let's merge it.

PR #190, opened 2022-11-29
Do not try to install the service worker at the root

juliandescottes:
The current example page for service workers is not working at https://mdn.github.io/dom-examples/service-worker/simple-service-worker/

The service worker is not installed because we attempt to register it for `https://mdn.github.io/` which is not allowed. Similarly all the URLs of the images point to `https://mdn.github.io/` which is wrong.


juliandescottes:
Anyone to review this? In its current form the example is not demonstrating service worker at all.
Maybe @teoli2003  ?

teoli2003 (approved):
LGTM (sorry for the delay)

PR #192, opened 2022-12-02
Improvements to `channel-messaging-multimessage`

MrBrain295:
Fix HTML checker warnings run Prettier and var -> const/let.


teoli2003 (approved):
👍

PR #195, opened 2022-12-28
Fix typo in code

ambujsahu81:


bsmth:
Thanks @ambujsahu81. The original is correct here so I'm going to close the PR.

The reason you need a backslash is because the message to log to the console is in single quotes:

```js
console.log('Your browser doesn\'t support web workers.'); 
```

You can try this yourself in the browser console:

image

If you want to avoid escaping the single quote character, you can enclose the string in double-quotes:

```js
console.log("Your browser doesn't support web workers."); 
```

I hope that helps. :)

ambujsahu81:
> 

Thanks for the explanation. :sweat_smile:

PR #198, opened 2023-01-16
Updating pointer lock demo

chrisdavidmills:
This PR updates the pointer lock demo according to the docs changes in https://github.com/mdn/content/pull/23100. Specifically:

* It adds usage of the `requestPointerLock()` option `unadjustedMovement`
* It cleans up and modernizes the code generally, e.g. by adding async/await, using `const`/`let`, and removing the browser-prefixed code that is no longer needed.
Once this is merged, I'll update the linked PR to reference this example.


chrisdavidmills:
@Elchi3 there we go.

Elchi3 (approved):
Thanks for doing this Chris! 👍

mustaqahmed:
We recently discovered that this change broke the MDN demo on Chrome on Linux plus any platform where "unadjusted movement" is not supported.

PR #210, opened 2023-03-08
Refactor target identifier logic

majidzeno:
### Description
The previous implementation used an if statement to check whether the event target had a first child or not before assigning it to the `targetIdentifier` variable. This was replaced with a simplified implementation that uses a logical OR operator to assign the first child or the target itself to `targetIdentifier`.

related https://github.com/mdn/content/pull/25149


wbamberg (approved):
👍 thank you @majidzeno !

PR #212, opened 2023-03-22
Add an example for HKDF

wbamberg:
See https://github.com/mdn/content/issues/25216.

This example is more or less a mashing together of the two existing examples, ECDH and PBKDF2.

We use ECDH to generate a shared secret, then HKDF to derive an AES encryption key from the shared secret.

I tried using `deriveKey` to get the HKDF key, rather than `deriveBits` followed by `importKey`. This worked fine on Chrome but not on Firefox, I don't know why.

************

You can try the example out at https://wbamberg.github.io/dom-examples/web-crypto/derive-key/index.html .

### DIFF
--- a/web-crypto/derive-key/hkdf.js
+++ b/web-crypto/derive-key/hkdf.js
@@ -1,28 +1,52 @@
 (() => {
-  let iv;
+  let salt;
   let ciphertext;
+  let iv;
+  let alicesSecretKey;
+  let bobsSecretKey;
 
   /*
-  Fetch the contents of the "message" textbox, and encode it
-  in a form we can use for the encrypt operation.
-  */
+    Fetch the contents of the "message" textbox, and encode it
+    in a form we can use for the encrypt operation.
+    */
   function getMessageEncoding() {
-    let message = document.querySelector("#ecdh-message").value;
+    let message = document.querySelector("#hkdf-message").value;
     let enc = new TextEncoder();
     return enc.encode(message);
   }
 
   /*
-  Encrypt the message using the secret key.
-  Update the "ciphertextValue" box with a representation of part of
-  the ciphertext.
+  Given some key material and some random salt,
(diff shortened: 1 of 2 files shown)


web-crypto/derive-key/hkdf.js
owlstead, line 1: In principle this is a nice example. I've got one nitpick: currently the IV is part of the fields of the instance. However, the IV should be communicated to the other party. This is a common mistake, and one that can lead to insecure code. Instead, the randomized IV should be prefixed to the ciphertext.

wbamberg: Thanks! Of course we're not actually sending anything in this example, but perhaps instead of a global `ciphertext` here we could have a `message` object with `iv` and `ciphertext` members, which might help emphasise that these things should be sent together?

Does the same apply to `salt`?

owlstead: It _could_ apply to salt. E.g. OpenSSL stores a salt when it derives a key from a password. It then derived _both the key and the IV_ from the salt - no need to randomize the ciphertext twice after all: the key / IV combination will only be used once if you do not reuse the salt.

OTOH, if you want to use the same derived key for multiple actions you'd probably have the salt stored somewhere else. HKDF actually works fine without a salt (i.e. empty salt) as well. For this example it makes sense to store the salt as well.

owlstead (changes requested):
IV and salt need to be stored as indicated, hasn't happened so far as I can see (?)

sideshowbarker:
@wbamberg I’m doing some triage of open mdn/dom-examples PRs, and noticed that this one seems to waiting on a response from you.

wbamberg:
@sideshowbarker , @owlstead , I finally got around to updating this PR!

sideshowbarker (approved):

PR #227, opened 2023-09-17
Improve navigator.share example

mrienstra:
Better error handling, was failing silently when not supported.

New version is similar to code example at https://developer.mozilla.org/en-US/docs/Web/API/Navigator/canShare#examples

This example is linked to from:
- https://developer.mozilla.org/en-US/docs/Web/API/Web_Share_API#example
- https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share#examples

(and possibly more, I didn't bother searching) and hosted at https://mdn.github.io/dom-examples/web-share/

If this PR is accepted, I'll open another that applies the same changes to the "live example" shown at https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share#examples


pepelsbey (approved):
Looks good! Thank you :) And sorry it took a while to review 😅

PR #230, opened 2023-09-30
Fix broken W3C Web Storage API link

iamericfletcher:
Replaced:
http://dev.w3.org/html5/webstorage/

With:
https://html.spec.whatwg.org/multipage/webstorage.html#storage


ferdnyc:
Good call, as the current link hits a 404 page.

wbamberg (approved):
👍 thank you!

PR #232, opened 2023-10-10
Web Storage demo: Remove webfonts

ferdnyc:
As the name of my PR branch implies, this was originally an effort to remove the demo's (hidden) use of Open Sans Condensed as its "sans-serif" font selection. The typeface is unnecessarily difficult-to-read and overstyled, distracting from the purpose of the demo.

But when I got into the code, I realized that it was forcing fonts for _all three_ style choices, while calling them simply "sans-serif", "serif", and "monospace".

There's no real reason to download web fonts and force them upon the user, just to present a simple API demo. While there's _**every**_ reason to keep the focus on the right thing: the technology being demonstrated. The best way to accomplish that is to eschew overdesigned presentation, and just rely on the user's (or operating system's) default font selections.


teoli2003 (approved):
Yes, let's keep these examples as simple as possible. The less distractions, the better.

teoli2003:
Thanks a lot!

This PR looks good to me.

Congratulations upon your first merged commit here! Welcome aboard! ✌️

ferdnyc:
MDN has been an invaluable resource, happy to be able to give back in some small way!

PR #234, opened 2023-10-24
fix(simple-service-worker): avoid HTML to appear as text

### LINKED ISSUE #191: Simple Service Worker example displays raw HTML
The doc https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers#service_workers_demo says https://bncb2v.csb.app/  should look like
image
but it looks like
image

Running Chrome Version 107.0.5304.110 (Official Build) (arm64) on macOS Ventura 13.0.1 (22A400)

Maximaximum:
Fixes https://github.com/mdn/dom-examples/issues/191


service-worker/simple-service-worker/app.js
```diff
@@ -42,7 +42,7 @@ const createGalleryFigure = async (galleryImage) => {
     const myImage = document.createElement('img');
     const myCaption = document.createElement('caption');
     const myFigure = document.createElement('figure');
-    myCaption.textContent = `${galleryImage.name}: Taken by ${galleryImage.credit}`;
+    myCaption.innerHTML = `${galleryImage.name}: Taken by ${galleryImage.credit}`;
```
caugner, line 45: Even though this works, we need to avoid [XSS](https://developer.mozilla.org/en-US/docs/Glossary/Cross-site_scripting). Instead of assigning the unescaped `galleryImage.name` via `innerHTML`, it would be better to split this up into two `span` elements, one of which is assigned via `textContent` and the other via `innerHTML`:

```suggestion
    const myName = document.createElement('span');
    myName.textContent = `${galleryImage.name}: `;
    const myCredit = document.createElement('span');
    myCredit.innerHTML = `Taken by ${galleryImage.credit}`;
    myCaption.append(myName, myCredit);
```

Ideally, the "Taken by " prefix would even move into the data.

caugner (changes requested):

pepelsbey (approved):
Looks good to me! Thank you both :)

caugner (approved):

PR #236, opened 2023-11-17
Add history api example

wbamberg:
@teoli2003 , the point of this is to replace https://developer.mozilla.org/en-US/docs/Web/API/History_API/Example, which I don't think is viable as an example.

I've used https://github.com/lesjames/history-api-demo, which in turn is based on http://html5doctor.com/history-api/, as a starting point, but have pretty much completely rewritten it.

The general idea is to show how to use the history API to allow back/forward navigation to work when you're doing a single-page app thing using fetch().

However I don't really know what I'm doing here so I'm very open to feedback!


history-api/script.js
```diff
@@ -0,0 +1,43 @@
+// Update the page with the new content
+function displayContent(content) {
+  const description = document.querySelector("#description");
```
teoli2003, line 3: Let's change the title too. That way, the history will be more useful:
Capture d’écran 2023-11-17 à 08 23 41


```suggestion
  document.title = `Creatures: ${content.name}`;
  
  const description = document.querySelector("#description");
```

history-api/script.js
```diff
@@ -0,0 +1,43 @@
...
+// Handle click on link
+document.addEventListener("click", async (event) => {
+  const creature = event.target.getAttribute("data-creature");
+  if (creature) {
+    event.preventDefault();
+    try {
+      const response = await fetch(`creatures/${creature}.json`);
+      const json = await response.json();
+      displayContent(json);
+      history.pushState(json, "", creature);
```
teoli2003, line 20: Adding a comment as this is the key point for the reader.
```suggestion
      // Add a new entry to the history.
      // This simulates loading a new page.
      history.pushState(json, "", creature);
```

history-api/script.js
```diff
@@ -0,0 +1,43 @@
...
+function displayContent(content) {
+  const description = document.querySelector("#description");
+  description.textContent = content.description;
+
+  const photo = document.querySelector("#photo");
+  photo.setAttribute("src", content.image.src);
+  photo.setAttribute("alt", content.image.alt);
+}
+
+// Handle click on link
```
teoli2003, line 11: NIT
```suggestion
// When a link is clicked: update the page for that creature.
// From a user perspective, this will look like the loading of a new page.

```

history-api/script.js
```diff
@@ -0,0 +1,43 @@
...
+    } catch (err) {
+      console.error(err);
+    }
+  }
+});
+
+// Handle forward/back buttons
+window.addEventListener("popstate", (event) => {
+  if (event.state) {
+    displayContent(event.state);
```
teoli2003, line 30:
```suggestion
   // Simulate the loading of the previous page
    displayContent(event.state);
```

history-api/script.js
```diff
@@ -0,0 +1,43 @@
...
+      history.pushState(json, "", creature);
+    } catch (err) {
+      console.error(err);
+    }
+  }
+});
+
+// Handle forward/back buttons
+window.addEventListener("popstate", (event) => {
+  if (event.state) {
```
teoli2003, line 29: Add a comment
```suggestion
  // If a state has been provided, we have a "simulated" page
  // and we update the current page.
  if (event.state) {
```

teoli2003 (changes requested):
This was really good.

- I propose to add a few comments, explaining the part that are related to the History API
- I've updated `document.title` in `displayContent()` so that the history in the navigator looks better.

teoli2003 (approved):
👍  Nice!

PR #237, opened 2023-11-17
fix: history-api demo

OnkarRuikar:
Some issues to fix:
- Considering the demo is hosted at https://mdn.github.io/dom-examples/history-api/ , the home page link is giving 404.
- Links are advertised as `*.html` but they reach non .html destinations.

/cc @wbamberg

### DIFF
--- a/history-api/script.js
+++ b/history-api/script.js
@@ -23,7 +23,7 @@ document.addEventListener("click", async (event) => {
       displayContent(json);
       // Add a new entry to the history.
       // This simulates loading a new page.
-      history.pushState(json, "", `${creature}.html`);
+      history.pushState(json, "", creature);
     } catch (err) {
       console.error(err);
     }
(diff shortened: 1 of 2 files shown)


history-api/script.js
```diff
@@ -23,7 +23,7 @@ document.addEventListener("click", async (event) => {
       displayContent(json);
       // Add a new entry to the history.
       // This simulates loading a new page.
-      history.pushState(json, "", creature);
+      history.pushState(json, "", `${creature}.html`);
```
wbamberg, line 26: When you say "advertised as" do you mean in the href? Do you think it is better to update the state, as here, or to update the href? (it feels better to me to update the href but not sure why).

OnkarRuikar: > When you say "advertised as" do you mean in the href?

Yes, when we hover over a link it shows the anchor's path in bottom left corner of the browser.

> it feels better to me to update the href but not sure why

In single page applications static looking links feel odd. Let's get rid of `.html` from paths.

wbamberg (approved):
👍 nice!

PR #239, opened 2023-11-23
Transfer SourceBuffer example from https://github.com/nickdesaulniers…

wbamberg:
This is a leftover bit of https://github.com/mdn/content/pull/30423, in which we move the example from https://github.com/nickdesaulniers/netfix/blob/gh-pages/demo/bufferAll.html to this repo, where we can maintain it.

The JS in this version matches that in the page at https://developer.mozilla.org/en-US/docs/Web/API/SourceBuffer#examples, I hope we haven't mangled it over the years.

See also https://github.com/nickdesaulniers/netfix/issues/10, where this transfer was discussed. Thanks @nickdesaulniers !


teoli2003 (approved):
👍

PR #241, opened 2023-12-20
Don't hardcode MIME type in web-dictaphone

### LINKED ISSUE #233: Error on Safari
Running the web-dictaphone on Safari Version 16.2 (18614.3.7.1.5) under macOS Ventura 13.1 (22C65) appears to work but the media player displays an error - see screenshot.

I don't think this is a local configuration issue?
Screenshot 2023-10-22 at 11 29 47

teoli2003:
Fixes #233

The problem was that the MIME type was not supported on Safari.

[f789f3f](https://github.com/mdn/dom-examples/commit/f789f3fafdd28705289514f7aa2213e98ff9c1a2) Instead of hardcoding it, I now use the MIME type of the `MediaRecorder` object, they will match on every browser.

I also cleaned comments and logs to match our most recent standard in [5f87061](https://github.com/mdn/dom-examples/pull/241/commits/5f8706157fbaa6f70826b6c549524a7f1fac5638).

I tested locally in Firefox and Safari, using a VSCode extension:

Capture d’écran 2023-12-20 à 11 27 38

### DIFF
--- a/media/web-dictaphone/scripts/app.js
+++ b/media/web-dictaphone/scripts/app.js
@@ -77,7 +77,7 @@ if (navigator.mediaDevices.getUserMedia) {
       soundClips.appendChild(clipContainer);
 
       audio.controls = true;
-      const blob = new Blob(chunks, { 'type' : 'audio/ogg; codecs=opus' });
+      const blob = new Blob(chunks, { 'type' : mediaRecorder.mimeType });
       chunks = [];
       const audioURL = window.URL.createObjectURL(blob);
       audio.src = audioURL;
--- a/media/web-dictaphone/scripts/app.js
+++ b/media/web-dictaphone/scripts/app.js
@@ -14,7 +14,7 @@ const canvasCtx = canvas.getContext("2d");
 
 // Main block for doing the audio recording
 if (navigator.mediaDevices.getUserMedia) {
-  console.log('The mediaDevices.getUserMedia() media is supported.');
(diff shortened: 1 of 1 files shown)


media/web-dictaphone/scripts/app.js
```diff
@@ -1,24 +1,20 @@
...
+// Visualiser setup - create web audio api context and canvas
 let audioCtx;
 const canvasCtx = canvas.getContext("2d");
 
-//main block for doing the audio recording
-
+// Main block for doing the audio recording
 if (navigator.mediaDevices.getUserMedia) {
-  console.log('getUserMedia supported.');
+  console.log('The mediaDevices.getUserMedia() media is supported.');
```
wbamberg, line 17: This comment is harder to understand than the one it's replacing. Should it be:

```suggestion
  console.log('The mediaDevices.getUserMedia() method is supported.');
```

? Also I wonder why we have single-quotes here, has this been Prettified?

wbamberg (changes requested):

media/web-dictaphone/scripts/app.js
teoli2003: I have not prettified it. Let's add a commit for this, then run Prettifier (I'm so used of Prettifier run automatically)

wbamberg (approved):
👍 thank you @teoli2003 !

PR #245, opened 2024-01-17
Add Shadow DOM examples

pepelsbey:
> ⚠️ The same demos were already merged in https://github.com/mdn/interactive-examples/pull/2707 but didn’t work as expected. Once these are merged, I’ll remove the previous ones.

### Description

Adds declarative shadow DOM examples that can only work with `{{EmbedGHLiveSample}}` macros as remote demos.

### Motivation

To support existing implementations in Chrome and Safari and for the upcoming one in Firefox 122 (behind the flag).

### Related issues and pull requests

- https://github.com/mdn/content/pull/31577
- https://github.com/mdn/content/pull/31490


bsmth (approved):
Thank you!

PR #246, opened 2024-01-20
Make sample compatible with older browers (specifically IE11)

JX75:
The sample was using ES6 features not relevant to the purpose of the sample :
- arrow functions
- `this kind of ${"string formatting"}`
- destructured assignation, i.e. [a,b] = [a+b,a]

I replaced the first two features with vintage JavaScript and the third one with a traditional hack that should convince people that destructured assignement is a valuable improvement of JS.

It now runs with IE11.


teoli2003:
Sorry, the web has moved forward. We don't care about IE anymore.

Modern JS has to be used: if you need backward-compatibility you should use a transpiler.

PR #248, opened 2024-01-31
New EditContext API demo

captainbrosset:
This new demo is for the sake of this new guide: https://github.com/mdn/content/pull/32027
It should be merged before the guide.

To learn more about the EditContext API, see: https://developer.mozilla.org/en-US/docs/Web/API/EditContext_API

### DIFF
--- a/edit-context/README.md
+++ b/edit-context/README.md
@@ -1,0 +1,9 @@
+# EditContext API demos
+
+This folder contains demos for the [EditContext API](https://developer.mozilla.org/docs/Web/API/EditContext_API).
+
+## Demos
+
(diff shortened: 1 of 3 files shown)


captainbrosset:
> This ought to add an link to the top-level README, so it's findable from there.

I will add it in a new PR today.

> Also the directory here (https://github.com/mdn/dom-examples/tree/main/edit-context) does not match the links in the [tutorial PR](https://github.com/mdn/content/pull/32027) (https://mdn.github.io/dom-examples/editcontext-api/html-editor/).

Yeah, I've fixed that today.

PR #249, opened 2024-02-01
EditContext: Handle backward selections and remove double underline

captainbrosset:
The EditContext API now supports backward selections, so I needed to update the code. Also, the double underline style for IME text formats is going away, so I removed the code for it.

### DIFF
--- a/edit-context/html-editor/index.html
+++ b/edit-context/html-editor/index.html
@@ -140,14 +140,6 @@
       text-decoration: underline 2px;
     }
 
-    ::highlight(ime-double-thin) {
-      text-decoration: underline double 1px;
-    }
-
-    ::highlight(ime-double-thick) {
-      text-decoration: underline double 2px;
-    }
-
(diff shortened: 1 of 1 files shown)


Rumyra (approved):

edit-context/html-editor/index.html
```diff
@@ -475,25 +460,31 @@
...
+          anchorNode: isBackwards ? extentNode : anchorNode,
+          anchorOffset: isBackwards ? extentOffset : anchorOffset,
+          extentNode: isBackwards ? anchorNode : extentNode,
+          extentOffset: isBackwards ? anchorOffset : extentOffset
+        };
       }
 
       // Render the initial view.
       render();
+      setInterval(render, 1000);
```
wbamberg, line 487: Can you explain why we have to do this now? It has the unfortunate effect of making it impossible to use the devtools inspector for the rendered text.

PR #251, opened 2024-02-01
Add the new edit-context demo to the root-level README.md file

captainbrosset:
This way, it can be found from the repo homepage.

Oh, also, I'm removing the `en-US` part from existing MDN URLs.


wbamberg (approved):
👍 thank you Patrick!

PR #252, opened 2024-02-02
Simplify the backwards selection code in the EditContext demo

captainbrosset:
As noted in https://github.com/mdn/content/pull/32027#discussion_r1475359746, the code to handle backwards selections in the EditContext demo can be simplified.


dandclark (approved):

wbamberg (approved):
👍

PR #254, opened 2024-02-03
service-worker: await unnecessary, errors should be handled

sidewayss:
Pretty simple stuff, minimal code changed.  `await` as the last command in a function is pointless.  Errors thrown during the completion of a promise should be handled.  Low probability of errors in this project as-is, but high probability of errors for any newbie adapting this example for their own project.


skyclouds2001 (commented):
The original one is intended to throw the exception to the outside where the function is called, whether its thrown exception should be handled or not completely decided by outside caller.
The fix make it outside caller unable to handle it any more.

PR #259, opened 2024-02-22
chore: Format CSS

bsmth:
Excluding the HTML & JS for the moment, we can format CSS in one pass here for consistency.


pepelsbey (approved):
Looks good to me. Thanks!

PR #262, opened 2024-02-27
chore: remove unused tutorial files

### LINKED ISSUE #201: dom-examples-main/webgl-examples/tutorial/tetrahedron not working
tetrahedron says gl-matrix.js and webgl.css don't exist.

bsmth:
These files are not referenced anywhere in content and the examples are broken. These should be removed from this repo.

Fixes #201


pepelsbey (approved):
It seems like this demo was [moved from the webgl-examples repo](https://github.com/mdn/webgl-examples/commit/3a03dcd26ac1ab40de36bdefbe47c7b2c50f9221) before it got archived. But then something happened with these two files. I guess there’s no use in keeping the demo if it’s not used anywhere.

PR #265, opened 2024-03-15
DO NOT MERGE YET: Add cross-document view transitions demo

chrisdavidmills:
> **Note:** DO NOT MERGE until after the corresponding content updates in https://github.com/mdn/content/pull/32723 have been merged, as this PR will break the demo links in the current content.

This PR adds a basic cross-document (MPA) [view transitions](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API) demo to the repo.

It reorganizes the repo a bit, so that the existing SPA view transitions demo is put in a separate `spa` folder, and the new one in an `mpa` folder.

Note that as of March 25th, these features only work in Chrome with chrome://flags/#view-transition-on-navigation flag enabled.

Related PRs:

- BCD: https://github.com/mdn/browser-compat-data/pull/22638
- MDN content update: https://github.com/mdn/content/pull/32723


README.md
```diff
@@ -73,7 +73,7 @@ Code examples that accompany various MDN DOM and Web API documentation pages.
 
 - The "web-storage" directory contains a basic demo to show usage of the Web Storage API. For more detail on how it works, read [Using the Web Storage API](https://developer.mozilla.org/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API). [View the demo live](https://mdn.github.io/dom-examples/web-storage/).
 
-- The "view-transitions" directory contains a basic demo to show usage of the [View Transitions API](https://developer.mozilla.org/docs/Web/API/View_Transitions_API). [View the demo live](https://mdn.github.io/dom-examples/view-transitions/).
+- The "view-transitions" directory contains demos to show usage of the [View Transitions API](https://developer.mozilla.org/docs/Web/API/View_Transitions_API), both for single-page app [View the SPA demo live](https://mdn.github.io/dom-examples/view-transitions/spa/) and multiple-page app [View the MPA demo live](https://mdn.github.io/dom-examples/view-transitions/mpa/) view transitions.
```
pepelsbey, line 76: I’d suggest adding a note to the MPA demo, something like:

> As of March 15, 2024 it works only in Chrome with `chrome://flags/#view-transition-on-navigation` flag enabled.

It took me some time to figure it out :)

view-transitions/mpa/index.js
```diff
@@ -0,0 +1,13 @@
...
+
+window.addEventListener("pageswap", event => {
+    console.log("pageswap fired");
+    // event.viewTransition.skipTransition();
+});
+
+
+window.addEventListener("pagereveal", event => {
+    console.log("pagereveal fired");
+    // event.viewTransition.skipTransition();
```
pepelsbey, line 11: It’s not clear what they do and why they’re commented. Given the name, they probably let you skip the transition. But a small note should make it more straightforward :)

view-transitions/mpa/styles.css
```diff
@@ -0,0 +1,73 @@
...
+@view-transition {
+    navigation: auto;
+}
+
+/* Customize the default animation */
+
+/* ::view-transition-old(root),
+::view-transition-new(root) {
+    animation-duration: 0.5s;
+} */
```
pepelsbey, line 42: I wonder if it would be better to enable this customization by default. This would invite you to play with values and explore the behavior right in the DevTools.

pepelsbey (commented):
Looks good to me. Just a few ideas how to improve it a bit :)

view-transitions/mpa/index.js
chrisdavidmills: Sorry, I didn't mean to include these. I'm going to remove the JS from the example altogether. I'll possibly include some JS demos separately as future updates.

view-transitions/mpa/styles.css
chrisdavidmills: Good idea; I've done this in my next commit.

chrisdavidmills:
@pepelsbey please can you approve this one so we can get it merged? Otherwise the demo links will be broken in the view transitions docs updates. Thanks!

pepelsbey (approved):
Thanks!

chrisdavidmills:
> Thanks!

@pepelsbey yay, cheers mate!

PR #268, opened 2024-03-22
fix(html): Make script elements children of head or body elements

### LINKED ISSUE #267: Code examples have stray <script> tags

A lot code examples in the `mdn/dom-examples` codebase with inline JavaScript have their `<script>` tag placed after the `</body>` close tag. Though the JavaScript is executed, this is technically invalid. The [HTML5 specification][] tells us that the body can only contain a `<head>` and `<body>` tags.

> ### 4.1.1 The `html` element [§][HTML5 specification]
> 
>   - **[Categories][]:**
>     None.
>   - **[Contexts in which this element can be used][]:**
>     As document's [document element][].
>     Wherever a subdocument fragment is allowed in a compound document.
>   - **[Content model][]:**
>     A [`head`][] element followed by a [`body`][] element.

When looking at the source of a page via `view-source:https://mdn.github.io/dom-examples/abort-api/` Firefox will also notify the reader that the tag isn't placed correctly.

![screenshot view-source Firefox][]

Validators like https://validator.w3.org will also yield the same error when validating the page: 

> Error: Stray start tag `script`.

Examples of misplacement of the `<script>` can be found plentiful within the `mdn/dom-examples` codebase. Here are some examples:

https://github.com/mdn/dom-examples/blob/929e39675d42b6642e96ede0b7cf4aa25eb822f5/abort-api/index.html#L39-L40
https://github.com/mdn/dom-examples/blob/929e39675d42b6642e96ede0b7cf4aa25eb822f5/fetch/basic-fetch/index.html#L17-L18
https://github.com/mdn/dom-examples/blob/929e39675d42b6642e96ede0b7cf4aa25eb822f5/fetch/fetch-response/index.html#L16-L17

[HTML5 specification]: https://html.spec.whatwg.org/multipage/semantics.html#the-html-element
[Categories]: https://html.spec.whatwg.org/multipage/dom.html#concept-element-categories
[Contexts in which this element can be used]: https://html.spec.whatwg.org/multipage/dom.html#concept-element-contexts
[document element]: https://dom.spec.whatwg.org/#document-element
[Content model]: https://html.spec.whatwg.org/multipage/dom.html#concept-element-content-model
[`head`]: https://html.spec.whatwg.org/multipage/semantics.html#the-head-element
[`body`]: https://html.spec.whatwg.org/multipage/sections.html#the-body-element
[screenshot view-source Firefox]: https://github.com/mdn/dom-examples/assets/1522964/dbf34064-836d-47ea-974b-6ecdd4f215d9

I would expect to see valid HTML, meaning that the `<script>` tags have to be placed before the closing `</body>` tag (or inside the `<head>`).

_No response_

_No response_

bsmth:
This PR fixes HTML validation errors relating to where the script element is permitted.

__Changes:__
* Prettier formatting in parallel in changed files
* `<script>` elements moved inside `<head>` or `<body>`

__Additions:__
* `html-validate` config - see https://html-validate.org/usage/index.html

There are two html-validate violations for the `element-permitted-content` rule left over after this PR:

```bash
html-validate ./**/*.html | grep -B 2 element-permitted-content
# dom-examples/popover-api/toggle-help-ui/index.html
#  24:6   error  <button> is missing recommended "type" attribute     no-implicit-button-type
#  37:10  error  <hr> element is not permitted as content under <ul>  element-permitted-content
#
# dom-examples/shadow-dom/shadowrootmode/scoping.html
#  16:8  error  <style> element is not permitted as content under <article>  element-permitted-content
```

## How to review this PR

To ignore whitespace-only changes, view files using the following view:

https://github.com/mdn/dom-examples/pull/268/files?diff=split&w=1

Fixes #267

### DIFF
--- a/insert-adjacent/insertAdjacentText.html
+++ b/insert-adjacent/insertAdjacentText.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01//EN">
+<!DOCTYPE html>
 <html lang="en">
   <head>
     <title>insertAdjacentText() demo</title>
--- a/insert-adjacent/insertAdjacentText.html
+++ b/insert-adjacent/insertAdjacentText.html
@@ -1,57 +1,55 @@
-<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN">
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01//EN">
 <html lang="en">
-<head>
-  <title>insertAdjacentText() demo</title>
-  <style>
-    section p {
-      font-size: 1.1rem;
-      color: blue;
-    }
+  <head>
+    <title>insertAdjacentText() demo</title>
+    <style>
+      section p {
+        font-size: 1.1rem;
+        color: blue;
+      }
 
-    input {
-      margin-bottom: 0.5rem;
-    }
-  </style>
-</head>
-<body>
+      input {
+        margin-bottom: 0.5rem;
+      }
+    </style>
+  </head>
+  <body>
+    <p>
+      Enter some text to add, then use the first two buttons below to insert
+      your text after the existing text.
+    </p>
 
-<p>Enter some text to add, then use the first two buttons below to insert your text after the existing text.</p>
+    <section>
+      <p>This is my text</p>
+    </section>
 
-<section>
-  <p>This is my text</p>
-</section>
+    <input type="text" /><br />
+    <button class="before">Insert before</button>
+    <button class="after">Insert after</button>
+    <button class="reset">Reset demo</button>
 
-<input type="text"><br>
-<button class="before">Insert before</button>
-<button class="after">Insert after</button>
-<button class="reset">Reset demo</button>
+    <script>
+      var beforeBtn = document.querySelector(".before");
+      var afterBtn = document.querySelector(".after");
+      var resetBtn = document.querySelector(".reset");
 
-</body>
+      var para = document.querySelector("section p");
+      var initContent = para.textContent;
 
-<script>
+      var textInput = document.querySelector("input");
 
-  var beforeBtn = document.querySelector('.before');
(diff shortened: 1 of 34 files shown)


insert-adjacent/insertAdjacentText.html
```diff
@@ -1,57 +1,55 @@
-<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN">
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01//EN">
```
pepelsbey, line 1:
```suggestion
<!DOCTYPE html>
```

matchmedia/index.html
```diff
@@ -1,60 +1,59 @@
-<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN">
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01//EN">
```
pepelsbey, line 1:
```suggestion
<!DOCTYPE html>
```

pepelsbey (approved):
While we’re at it, I’d suggest replacing ancient doctypes with modern ones. Fortunately, they’re identical from the browser parsers’ perspective. Otherwise, LGTM :) Thanks!

insert-adjacent/insertAdjacentText.html
bsmth: Good catch

bsmth:
> Oh, BTW, [https://html-validate.org/](https://html-validate.org/usage/index.html) is not an official validator for the HTML spec. They use their own set of rules, parser, etc. I’d suggest using https://validator.w3.org/ and packages based on it. For example, I use [w3c-html-validator](https://www.npmjs.com/package/w3c-html-validator). It’s not “offline”, but it uses the W3C’s API for checking the document.

Good call, removed in https://github.com/mdn/dom-examples/pull/268/commits/378f31b0a4a6c8b26e61c9ee6c044d1fb51e2e66

Going to merge shortly, ty!

PR #269, opened 2024-04-01
MessageChannel-postMessage-error-handle

polikm5:
If there is a special plugin in chrome that listens for postMessage, it will throw an error
image
image


pepelsbey (approved):
Nice catch! Thank you :)

PR #271, opened 2024-04-23
chore: remove example with no content references

bsmth:
Deleting this one as it looks like a duplicate of fetch with init object. There are no references to it in content since https://github.com/mdn/content/pull/33214


pepelsbey (approved):
Good catch! Thank you :)

bsmth:
Thank you!

PR #274, opened 2024-05-08
Add early return to document picture-in-picture example to match docs

joshfarrant:
- Ran prettier on the modified file
- Updated `togglePictureInPicture` function to add an early return as discussed here https://github.com/mdn/content/pull/33293


pepelsbey (approved):
Looks good! Thank you :)

PR #277, opened 2024-07-23
fix: remove extra semicolons in web-share example

joshkadis:
@pepelsbey @mrienstra I noticed a small bug that was introduced in #227. There are a couple of places where a semicolon comes after a function argument, which makes the code invalid. Easiest to show from linting:

Screenshot 2024-07-23 at 3 05 56 PM


pepelsbey (approved):
Oh, good catch! Thank you :)

PR #280, opened 2024-08-27
Small tweak to visual viewport example wording

chrisdavidmills:
This PR updates the text in the HTML of the visual viewport example to match the same text change made in the example explanation in the docs (see https://github.com/mdn/content/pull/34427).


wbamberg (approved):
👍

PR #286, opened 2024-11-30
Click n drag

allan-bonadio:
This is a click-drag-drop example, using regular mouse events, not d&d events.  I was surprised that MDN didn't seem to have such an example yet!  I've been doing interaction like this for 10+ years, and I've picked up lots of unobvious tips and tricks, in many cases, the hard way.

I've been paying attention to the fact that other programmers will copy/paste into their own code, trying to make the code general, with lots of comments.

I've also added a set of suggestions for more modern methods and user-friendly effects.  This'll work on all browsers, as is, from the last 10 years.


allan-bonadio:
my rationale for this PR is that I once had a coworker whose task was to make a list re-arrangeable by clicking and dragging elements up or down.  She used the full d&d API, which was way overkill.  She had to design some code-independent data format, etc.  Despite the fact that the data was in the same address space, was involved with the UI in that page, and limited to its enclosing element.  And was limited to the associated JS code and the backend managing it.  

d&d is for dragging to and from other apps, or at least other web pages, not within the same JS global space.  These should be done with old fashioned mousedown/move/up events, as this example illustrates.  There's probably a lot of other programmers who have fallen into this trap.  Ultimately, I want the d&d article to point this out, but there's some more contributions I need to do first, after this and the  'mouse & pointer event pages updates #36993' PR .

bsmth:
Hi @allan-bonadio - thanks a lot for opening this one. I appreciate that a lot of work went into this, but I'm reluctant to add it instead of recommending the https://github.com/mdn/dom-examples/tree/main/drag-and-drop examples. Basically, we're targeting specialized browser APIs in this repo instead of workarounds that emulate them - these are easier (more consistent and predictable) for developers to debug, and have improvements land in them through standards work, etc. I'm sorry if that's disappointing, but thanks again for the suggestions!

allan-bonadio:
But, this is completely different.  D&D goes through the operating system.  It's designed to drag data to and from software external to the browser, not stuff within the same page.  In fact, the MDN pages about D&D are all confused because starting a D&D and receiving a drop are completely independent operations.  You can do either without the other.  But the MDN D&D pages throw all the events into one big jumble, and there's no suggestion that the two halves are independent.

Think of a list of items somewhere in your UI.  You want the user to be able to click and drag to rearrange them.  There would be no reason to drag an item outside of the list itself - let alone to a text editor or file manager.  There's no actual data being moved; just a rearranging of some array internal to the UI.  Why should a programmer have to come up with a MIME type for the data - which doesn't exist?  Why should there be a 'data transfer' object?  Why should the OS be involved?  D&D has all this additional machinery that adds complexity, but no features, to rearranging a list.

Now, think of an SVG display of a triangle (or any polygon).  Each corner of the triangle has a dot, and the user can click and drag the corner dots to other locations.  As they do, the triangle reshapes to follow.  So the Move events simply move the vertex of the triangle.  In this example, there's no data moving anywhere; it's just a UI device to get data from the user.  

This is exactly what mouseDown/Up, mouseMove and mouseEnter/Leave events were designed for, decades ago.  I've been pissed for years that MDN has almost no documentation describing this widely used API.  The closest is this lame demo where the user draws a line on a canvas.  There isn't even any animation to it.  There's a lot of subtleties and gotchas; MDN should be telling people about these.

People  make their own widgets with this api: dropdown menus, drawers that open from the bottom of a panel, dividers that the user can slide back and forth between two areas.  These are all done with the raw mouse events.

PR #291, opened 2025-01-29
fix audio source scope to enable stopping playback

AKR-2803:
- **Issue:** The audio source [created here](https://github.com/mdn/dom-examples/blob/085b5952ccc3171b1075c4027868b191a745d3cf/fetch/fetch-array-buffer/index.html#L44) is accessible to the `play()` function, but **inaccessible** to the `stop()` function. As a result, the `stop` button **fails to stop** the audio playback.
- **Fix:** Moved audio source to a globally accessible variable (`currentSource`), allowing the `stop()` function to access it and stop the playback. Added a check to ensure `stop()` is only called when a valid source exists.


pepelsbey (approved):
Oh, good catch! Thank you :)

PR #293, opened 2025-01-31
Update image path to folder nested under dom-examples

wesgarrison:
Old: https://mdn.github.io/to-do-notifications/img/icon-128.png
New: https://mdn.github.io/dom-examples/to-do-notifications/img/icon-128.png

### DIFF
--- a/to-do-notifications/scripts/todo.js
+++ b/to-do-notifications/scripts/todo.js
@@ -309,7 +309,7 @@ window.onload = () => {
   // Create a notification with the given title
   function createNotification(title) {
     // Create and show the notification
-    const img = '/to-do-notifications/img/icon-128.png';
+    const img = '/dom-examples/to-do-notifications/img/icon-128.png';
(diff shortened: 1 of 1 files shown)


to-do-notifications/scripts/todo.js
```diff
@@ -309,7 +309,7 @@ window.onload = () => {
   // Create a notification with the given title
   function createNotification(title) {
     // Create and show the notification
-    const img = '/to-do-notifications/img/icon-128.png';
+    const img = '/dom-examples/to-do-notifications/img/icon-128.png';
```
bsmth, line 312: This is good for the GitHub version, but will break on local ones, should we do something like:

```js
const basePath = window.origin.includes("github.io") 
  ? "/dom-examples" 
  : "";

const img = `${basePath}/to-do-notifications/img/icon-128.png`;
```

If it's on GitHub, add the repo to the base path, otherwise use an empty string?

PR #294, opened 2025-02-05
Remove logging code from screen capture demos

chrisdavidmills:
In his [review of my screen capture extensions docs](https://github.com/mdn/content/pull/36939), @wbamberg raised the point that the logging code in the demos is not really adding anything. I agree, so this PR removes the logging code from those demos.


bsmth (approved):
LGTM, thanks 👍🏻

PR #295, opened 2025-02-06
Add popover=hint example

chrisdavidmills:
This PR adds a demo to demonstrate typical usage of the `popover="hint"` attribute value, which is supported in Chrome 133.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Code examples that accompany various MDN DOM and Web API documentation pages.
 
 - The "pointer-lock" directory contains a basic demo to show usage of the Pointer Lock API. You can find more explanation of how the demo works at the main MDN [Pointer Lock API](https://developer.mozilla.org/docs/Web/API/Pointer_Lock_API) page. [Run the demo live](https://mdn.github.io/dom-examples/pointer-lock/).
 
-- The "popover-api" directory is for examples and demos of the [Popover API](https://developer.mozilla.org/docs/Web/API/Popover_API) standard. Go to the [Popover API demo index](https://mdn.github.io/dom-examples/popoover-api/) to see what's available.
+- The "popover-api" directory is for examples and demos of the [Popover API](https://developer.mozilla.org/docs/Web/API/Popover_API) standard. Go to the [Popover API demo index](https://mdn.github.io/dom-examples/popover-api/) to see what's available.
 
 - The "reporting-api" directory contains a couple of basic demos to show usage of the Reporting API. You can find more explanation of how the API works in the main MDN [Reporting API](https://developer.mozilla.org/docs/Web/API/Reporting_API) docs. [Run the deprecation report demo live](https://mdn.github.io/dom-examples/reporting-api/deprecation_report.html).
 
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Code examples that accompany various MDN DOM and Web API documentation pages.
 
 - The "pointer-lock" directory contains a basic demo to show usage of the Pointer Lock API. You can find more explanation of how the demo works at the main MDN [Pointer Lock API](https://developer.mozilla.org/docs/Web/API/Pointer_Lock_API) page. [Run the demo live](https://mdn.github.io/dom-examples/pointer-lock/).
 
-- The "popover-api" directory is for examples and demos of the [Popover API](https://developer.mozilla.org/docs/Web/API/Popover_API) standard. Go to the [Popover API demo index](popover-api/) to see what's available.
+- The "popover-api" directory is for examples and demos of the [Popover API](https://developer.mozilla.org/docs/Web/API/Popover_API) standard. Go to the [Popover API demo index](https://mdn.github.io/dom-examples/popoover-api/) to see what's available.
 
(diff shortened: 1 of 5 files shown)


popover-api/popover-hint/index.css
```diff
@@ -0,0 +1,148 @@
...
+[popover="auto"] button:focus {
+  background: #ccc;
+}
+
+[popover="auto"] button:active {
+  box-shadow: inset 1px 1px 2px #333, inset -1px -1px 2px #eee;
+}
+
+#submenu-1 {
+  position-anchor: --menu-anchor-1;
```
mfreed7, line 110: You shouldn't need to add `position-anchor` or `anchor-name` to anything in these demos, if you just add `showPopover({source: invoker})` to the JS bit. Declarative `popovertarget` activation automatically creates an implicit anchor reference, as does using `source` in `showPopover()`.

mfreed7 (approved):
Demo looks good to me!

popover-api/popover-hint/index.css
chrisdavidmills: @mfreed7 Nice! That works like a charm. Updated.

QQ: We don't actually have the `show` option documented on MDN. I'm going to add it in the PR associated with this demo, but I should also add a compat data entry for it at some point. When was `show` first supported?

chrisdavidmills:
@bsmth are you OK to have a look at this one? Alternatively, @pepelsbey might enjoy reviewing it ;-)

popover-api/popover-hint/index.html
```diff
@@ -0,0 +1,50 @@
...
+      <button>Option A</button><br /><button>Option B</button>
+    </div>
+    <div id="submenu-3" popover="auto">
+      <button>Option A</button><br /><button>Option B</button>
+    </div>
+
+    <div id="tooltip-1" class="tooltip" popover="hint">Tooltip A</div>
+    <div id="tooltip-2" class="tooltip" popover="hint">Tooltip B</div>
+    <div id="tooltip-3" class="tooltip" popover="hint">Tooltip C</div>
+  </body>
```
bsmth, line 49: Can we add something here like:

https://github.com/mdn/dom-examples/blob/8884ad4ba7043fa9d15b05d38b3e15836f81d4bd/popover-api/toggle-help-ui/index.html#L26

Saying what to look out for? Along the lines of 'mouseover the buttons to see effect of `popover="hint"`, clicking shows & hides the `popover="auto"` ones with a light dismiss of the `popover="hint"` ones'?

chrisdavidmills: I've added a help paragraph in my latest commit. I didn't want to try to say too much about the behavior as it is quite complex and hard to explain in a small space. Instead, I've added a general description of what is happening, and linked to the MDN explanation (which will be published soon) for more details.

popover-api/popover-hint/index.css
```diff
@@ -0,0 +1,112 @@
...
+[popover="auto"] button:active {
+  box-shadow: inset 1px 1px 2px #333, inset -1px -1px 2px #eee;
+}
+
+/* Styling hint popovers */
+
+[popover="hint"] {
+  inset: unset;
+  position: absolute;
+  top: calc(anchor(bottom) + 5px);
```
bsmth, line 103: Only issue I have with this is that it looks weird on browsers that don't support it, you get a tooltip off the bottom of the page and some scrollbars on mouseover. It looks like it's hard to do proper feature detection on that level of granularity (specifically ~`supports popover = hint`), isn't it?

popover-api/popover-hint/index.html
bsmth: Yeah that's great, tnx 👍🏻

popover-api/popover-hint/index.css
chrisdavidmills: You could feature detect based on the popover property of the DOM element, but it'd be a right pain building an alternative that still provides an acceptable experience in non-supporting browsers. What do think is needed here?

chrisdavidmills: OK, I've added in a note about browser support, and added in a media query to stop the note box from covering the button bar on narrow screens

popover-api/popover-hint/index.css
mfreed7: Great! Yeah, I think that's an outstanding item: https://github.com/mdn/mdn/issues/598. It's been shipped in Chromium since M133.

README.md
```diff
@@ -51,7 +51,7 @@ Code examples that accompany various MDN DOM and Web API documentation pages.
 
 - The "pointer-lock" directory contains a basic demo to show usage of the Pointer Lock API. You can find more explanation of how the demo works at the main MDN [Pointer Lock API](https://developer.mozilla.org/docs/Web/API/Pointer_Lock_API) page. [Run the demo live](https://mdn.github.io/dom-examples/pointer-lock/).
 
-- The "popover-api" directory is for examples and demos of the [Popover API](https://developer.mozilla.org/docs/Web/API/Popover_API) standard. Go to the [Popover API demo index](popover-api/) to see what's available.
+- The "popover-api" directory is for examples and demos of the [Popover API](https://developer.mozilla.org/docs/Web/API/Popover_API) standard. Go to the [Popover API demo index](https://mdn.github.io/dom-examples/popoover-api/) to see what's available.
```
bsmth, line 54: Typo:

```suggestion
- The "popover-api" directory is for examples and demos of the [Popover API](https://developer.mozilla.org/docs/Web/API/Popover_API) standard. Go to the [Popover API demo index](https://mdn.github.io/dom-examples/popover-api/) to see what's available.
```

bsmth (approved):
Thanks, Chris!

PR #297, opened 2025-02-27
Add FileSystemObserver demo

chrisdavidmills:
This PR adds a `FileSystemObserver` example to the repo, as per the conversation at https://github.com/mdn/content/pull/38270#discussion_r1968619081 and follow-up at https://github.com/mdn/content/pull/38270#issuecomment-2687476888

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,8 @@ Code examples that accompany various MDN DOM and Web API documentation pages.
 
 - The "edit-context" directory contains examples demonstrating the [EditContext API](https://developer.mozilla.org/docs/Web/API/EditContext_API). See the [list of examples](https://github.com/mdn/dom-examples/tree/main/edit-context/).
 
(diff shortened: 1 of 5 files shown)


bsmth (approved):
Thanks!

chrisdavidmills:
> On Firefox, https://github.com/chrisdavidmills/dom-examples/blob/6729dc1a0d1a050a64add4b364ff8fceb14dca69/filesystemobserver/random.js#L3 throws with:
> 
> > Uncaught (in promise) TypeError: rootHandle.remove is not a function
> 
> ...so the code at https://github.com/chrisdavidmills/dom-examples/blob/6729dc1a0d1a050a64add4b364ff8fceb14dca69/filesystemobserver/random.js#L11-L13 doesn't get to run.

Yup. This is because the [`remove()`](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemHandle/remove#browser_compatibility) method is not supported outside Chromium. But then again, neither is `FileSystemObserver`.

wbamberg:
I've filed this as https://github.com/mdn/dom-examples/issues/298.

PR #300, opened 2025-03-03
Fix broken non-support behavior in FileSystemObserver example

### LINKED ISSUE #298: FileSystemObserver demo doesn't work properly on non-Chromium browsers

-> https://github.com/mdn/dom-examples/pull/297#issuecomment-2688723535

The FileSystemObserver demo wants display a "not supported" message on browsers that don't support that API (i.e. non-Chromium browsers), and has a test for this, but it doesn't work, because before the test the demo calls a different API that's not supported in Chromium. So the helpful message never gets shown.

the message to be shown

no

_No response_

chrisdavidmills:
As explained in https://github.com/mdn/dom-examples/issues/298, the `FileSystemObserver` demo is meant to fail gracefully on browsers that don't support `FileSystemObserver` with a helpful message.

However, the message never gets shown on non-supporting browsers like Firefox and Safari because they also don't support the `remove()` method, which gets run outside the `FileSystemObserver` test structure.

This PR fixes the problem by moving the line that invokes `remove()` inside the test structure as well.

Fixes https://github.com/mdn/dom-examples/issues/298

### DIFF
--- a/filesystemobserver/random.js
+++ b/filesystemobserver/random.js
@@ -1,10 +1,10 @@
 (async () => {
   const rootHandle = await navigator.storage.getDirectory();
-  await rootHandle.remove({ recursive: true });
   await startObserver();
 
   async function startObserver() {
     if ("FileSystemObserver" in self) {
+      await rootHandle.remove({ recursive: true });
       await import("./observer.js");
       demoOPFS();
     } else {


bsmth (approved):
LGTM in Fx 135.0.1 (and working as expected in Chrome)

PR #302, opened 2025-03-21
Add CSS carousel examples

chrisdavidmills:
I am in the process of documenting the new CSS Carousels functionality on MDN (see https://github.com/mdn/content/pull/38752).

I have included the examples as live examples in the guide, but I also want to link to standalone live examples and source code, as they are not as easy to follow in-page.


pepelsbey (changes requested):
Looks good! Though I found an issue with viewport units and scroll bars. It’s better to avoid using `100vw` at any cost, since it’s the **viewport** width, not the “window content apart from scrollbars” width. And the moment you have vertical scrollbars, you’ll get this:

Image

You can check this by setting your scrollbars to be visible by default, not overlay content:

Image

So I would suggest using 100% instead of 100vw. Unfortunately, with the full-width panels, it doesn’t play nicely with flexbox, so I used a grid layout instead:

```
display: grid;
grid-auto-flow: column;
grid-auto-columns: 100%;
gap: 4%;
```

Image

pepelsbey (approved):
Thanks!

PR #306, opened 2025-05-14
Use keydown instead of the deprecated keypress

wbamberg:
Sync with https://github.com/mdn/content/pull/8862, see also https://github.com/mdn/content/pull/39506.


dipikabh (approved):
Thank you!

PR #312, opened 2025-06-26
Add CSS progress() function demo

chrisdavidmills:
This PR adds a demo for the CSS progress() function.

I'm going to link to the demo in this case rather than include it on the page as a live example, because it needs a wide screen to really show off the effect I'm creating.

The reference `shared-assets` image is added in https://github.com/mdn/shared-assets/pull/52.


dipikabh (approved):
Wow! Looks like a fantastic effect (Chrome 138).
My only comment is that the text is blending with the background image. You might want to add a solid or semi-transparent background color to improve its legibility.

PR #316, opened 2025-07-03
Add Launch Handler API demo

chrisdavidmills:
The MDN [Launch Handler API documentation](https://developer.mozilla.org/en-US/docs/Web/API/Launch_Handler_API) has several pages that link to [this demo hosted on Glitch](https://launch-handler.glitch.me/).

Because Glitch is going away, this PR re-hosts the demo on the `dom-examples` repo.


tomayac (approved):
LGTM, this seems like a direct port-over of my Glitch demo.

bsmth (approved):
Leaving a +1, too 👍🏻

Thanks!

PR #319, opened 2025-07-03
Add contact picker api demo

chrisdavidmills:
The MDN [Contact Picker API documentation](https://developer.mozilla.org/en-US/docs/Web/API/Contact_Picker_API) has several pages that link to [this demo hosted on Glitch](https://contact-picker.glitch.me/).

Because Glitch is going away, this PR re-hosts the demo on the `dom-examples` repo.

Note that I've gotten rid of the common.js https checker code, as it seemed over the top to have JavaScript included that blocks the page from loading if it is not loaded over HTTPS. We're going to have it on HTTPS anyway. It also made my local environment go funny. I had to restart my testing server and Chrome before I was able to test any demos in the repo again.


tomayac (approved):
LGTM, this one is from @petele originally.

bsmth (approved):
Nice, thank you

PR #324, opened 2025-07-23
Internalize execCommand() editor example

Josh-Cena:
Part of mdn/content#16120. This example is interesting but too long, and uses bootstrap, so it's better to be standalone.

cc @bsmth

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto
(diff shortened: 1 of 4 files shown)


Josh-Cena:
Thanks! I keep messing up the line endings because I'm not in the habit yet and I'm soon going back to mac anyway.

bsmth (approved):
Great, thanks 👍🏻

PR #325, opened 2025-07-23
Add viewport segments api demo

chrisdavidmills:
Chrome 138 adds the Viewport Segments API, which provides a way to access the position and dimensions of logically separate viewport regions. Viewport segments are created when the viewport is split by one or more hardware features (such as a fold or a hinge between separate displays).

This PR adds an example to demonstrate how to use it.

### DIFF
--- a/viewport-segments-api/index.css
+++ b/viewport-segments-api/index.css
@@ -116,7 +116,7 @@ h1 {
 
 /* Media queries */
 
-@media (horizontal-viewport-segments: 1) and (vertical-viewport-segments: 1) and (orientation: landscape) {
+@media (orientation: landscape) {
   .wrapper {
     flex-direction: row;
   }
@@ -132,7 +132,7 @@ h1 {
   }
 }
 
-@media (horizontal-viewport-segments: 1) and (vertical-viewport-segments: 1) and (orientation: portrait) {
+@media (orientation: portrait) {
   .wrapper {
     flex-direction: column;
   }
--- a/viewport-segments-api/index.css
+++ b/viewport-segments-api/index.css
@@ -2,7 +2,6 @@
 
 html {
   font-family: helvetica, arial, sans-serif;
-  box-sizing: border-box;
   height: 100%;
 }
 
--- a/viewport-segments-api/index.css
+++ b/viewport-segments-api/index.css
@@ -177,7 +177,7 @@ h1 {
   }
 
   .list-view {
-    flex: 0 0 env(viewport-segment-height 0 0);
+    height: env(viewport-segment-height 0 0);
   }
 
   .fold {
@@ -189,6 +189,6 @@ h1 {
   }
 
   .detail-view {
-    flex: 0 0 env(viewport-segment-height 0 1);
+    height: env(viewport-segment-height 0 1);
   }
 }
--- a/viewport-segments-api/index.css
+++ b/viewport-segments-api/index.css
@@ -1,0 +1,194 @@
+/* Document styles */
+
+html {
+  font-family: helvetica, arial, sans-serif;
+  box-sizing: border-box;
+  height: 100%;
+}
+
+body {
+  margin: 0;
+  height: 100%;
+  background: #ffc8dd;
+}
+
+.wrapper {
+  height: 100%;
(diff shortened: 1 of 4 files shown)


chrisdavidmills:
@darktears, would you be able to help me test this demo? I adapted the viewport segments CSS from one of your demos, and didn't change it that much.

darktears:
The demo doesn't look quite right on my Z Fold 4. But I can see similar issues on the DevTools when I use the Device Emulation mode and pick the Z Fold 5 or the Asus Zenbook Fold.

viewport-segments-api/index.css
```diff
@@ -0,0 +1,194 @@
...
+  font-size: 1.4rem;
+}
+
+.segment-output p {
+  margin: 10px 0 0 0;
+}
+
+/* Media queries */
+
+@media (horizontal-viewport-segments: 1) and (orientation: landscape) {
```
darktears: I don't think you should mix this with the orientation. If there is a single segment it's just the regular normal case, just use the orientation property.

viewport-segments-api/index.js
```diff
@@ -0,0 +1,54 @@
...
+reportPostureOutput();
+navigator.devicePosture.addEventListener("change", reportPostureOutput);
+
+// Output viewport segment details to each segment
+
+const wrapperElem = document.querySelector(".wrapper");
+const listViewElem = document.querySelector(".list-view");
+const detailViewElem = document.querySelector(".detail-view");
+
+function addSegmentOutput(segments, index, elem) {
```
darktears: So, you want this to be executed whenever the configuration of the viewport change (this explains the bug where the panel doesn't show consistently). You need to listen to resize event (when the device is rotated) and update the value when the posture is changed but the device isn't rotated.

viewport-segments-api/index.css
```diff
@@ -116,7 +116,7 @@ h1 {
 
 /* Media queries */
 
-@media (horizontal-viewport-segments: 1) and (orientation: landscape) {
+@media (horizontal-viewport-segments: 1) and (vertical-viewport-segments: 1) and (orientation: landscape) {
```
darktears, line 119: Just remove the MQ matching horizontal/vertical-viewport-segments: 1 it's pointless.

chrisdavidmills: Done.

viewport-segments-api/index.js
darktears: I meant `window.addEventListener('resize', reportSegments);` and as you correctly wrote `navigator.devicePosture.addEventListener("change", reportSegments);`

Listening for the orientation is not a good way because for example on the Asus Zenbook Fold undocking/docking the keyboard will not change the orientation but the viewport will be resized.

chrisdavidmills: OK, so the `window.screen.orientation.addEventListener("change", reportSegments);` line is not needed?

What about when you rotate a device such as a smartphone, to change its orientation? does the `resize` handler pick that up? I wasn't sure if it would, as the screen area stays the same...but I guess, if the height and width are changing?

chrisdavidmills:
@pepelsbey this is ready for a look, if you fancy it ;-)

pepelsbey (changes requested):
Looks good! I only found an edge case with JS: when you click in between the links, the section on the right gets filled with all the content. I’d suggest narrowing down the event listener and also preventing the default behavior so that the links won’t add `#` to URL.

Image

viewport-segments-api/index.js
```diff
@@ -0,0 +1,54 @@
+const list = document.querySelector("ul");
+const detailHeading = document.querySelector(".detail-view h1");
+const postureOutput = document.querySelector(".posture-output");
+
+// Update detail view when list item is clicked on
+
+list.addEventListener("click", (event) => {
+  detailHeading.textContent = event.target.textContent;
+});
```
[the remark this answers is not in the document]
chrisdavidmills: Ah ha - good spot!

pepelsbey (approved):
Perfect :) Thank you!

Yttuf (approved):

PR #326, opened 2025-08-08
Add websockets example

wbamberg:
This adds a minimal websocket client/server to dom-examples.

For background about why, see https://github.com/orgs/mdn/discussions/817 - the [existing guide](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_client_applications) refers to the obsolete example at https://github.com/mdn/samples-server/tree/master/s/websocket-chat.

If this PR is merged, I'd like to update that guide to refer to this example.

The whole thing is a lightly edited version of the [existing Deno example](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_a_WebSocket_server_in_JavaScript_Deno).

I've updated it to be [bfcache-compatible](https://web.dev/articles/bfcache#close-open-connections), and I've verified that it is, using the Chrome devtools.

### DIFF
--- a/websockets/README.md
+++ b/websockets/README.md
@@ -2,7 +2,7 @@
 
 This directory contains a minimal example [WebSocket](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API) client and server.
 
-When the page is loaded, it creates a WebSocket connection to the server, then then sends a ping every second. The server listens for the ping and and a response. The client listens for the responses and logs them.
+When the page is loaded, it creates a WebSocket connection to the server, then sends a ping every second. The server listens for the ping and sends a response. The client listens for the responses and logs them.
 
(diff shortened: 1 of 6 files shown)


websockets/README.md
```diff
@@ -0,0 +1,17 @@
+# README
+
+This directory contains a minimal example [WebSocket](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API) client and server.
+
+When the page is loaded, it creates a WebSocket connection to the server, then then sends a ping every second. The server listens for the ping and and a response. The client listens for the responses and logs them.
```
bsmth, line 5:
```suggestion
When the page is loaded, it creates a WebSocket connection to the server, then sends a ping every second. The server listens for the ping and sends a response. The client listens for the responses and logs them.
```

bsmth (approved):
Some typos for you, but otherwise works for me after following the README 👍🏻

wbamberg:
Thank you!

PR #327, opened 2025-08-11
Add CSS custom functions examples

chrisdavidmills:
I'm about to start documenting [CSS custom functions](https://drafts.csswg.org/css-mixins-1/#defining-custom-functions), which has [landed in Chrome 139](https://chromestatus.com/feature/5179721933651968).

This PR adds some associated demos to the `dom-examples` repo.

### DIFF
--- a/css-custom-functions/color-adjust-functions/style.css
+++ b/css-custom-functions/color-adjust-functions/style.css
@@ -1,0 +1,50 @@
+/* Borrowed from Miriam Suzanne's custom CSS functions post
+   https://www.oddbird.net/2025/04/11/custom-functions/  */
+@function --transparent(--color <color>, --alpha <number>: 0.8) {
+  result: oklch(from var(--color) l c h / var(--alpha));
+}
+
+@function --lighter(--color <color>, --lightness-adjust <number>: 0.2) {
+  result: oklch(from var(--color) calc(l + var(--lightness-adjust)) c h);
+}
+
+@function --darker(--color <color>, --lightness-adjust <number>: 0.2) {
+  result: oklch(from var(--color) calc(l - var(--lightness-adjust)) c h);
+}
+
+html {
+  font-family: Arial, Helvetica, sans-serif;
+  height: 100%;
+}
+
+body {
+  margin: 0;
+  height: inherit;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  background-image: repeating-linear-gradient(
+    -45deg,
+    transparent 0 20px,
+    black 20px 40px
+  );
+}
+
+section {
+  width: 50%;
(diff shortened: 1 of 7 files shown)


css-custom-functions/color-adjust-functions/style.css
```diff
@@ -0,0 +1,52 @@
...
+body {
+  margin: 0;
+  height: inherit;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  background-image: repeating-linear-gradient(
+    -45deg,
+    transparent 0 20px,
+    black 20px 40px
```
pepelsbey, line 31:
```suggestion
    lightgrey 20px 40px
```

I would suggest using a less harsh gradient. This one did hurt my eyes a little :)

Image

css-custom-functions/responsive-narrow-wide/style.css
```diff
@@ -0,0 +1,29 @@
...
+@function --narrow-wide(--narrow, --wide) {
+  result: if(media(width < 700px): var(--narrow) ; else: var(--wide));
+}
+*/
+
+body {
+  display: grid;
+  grid-template-columns: repeat(--narrow-wide(1, 3), 1fr);
+  gap: --narrow-wide(0, 20px);
+  padding: 0 20px;
```
pepelsbey, line 19:
```suggestion
  padding: 0 20px;
  font-family: system-ui;
```

Unlike the other demos, you haven’t used the `Helvetica, Arial, sans-serif` spell here. Have you considered using this modern generic font-family value? It’s been Baseline widely available for a while now. It would help the demo look “styled”, you know.

pepelsbey (commented):
I like the demos! Thanks :) A few suggestions are below, along with an idea.

It might take a while for this feature to appear in other browsers (a year or two). So I wonder if we could use a simple “your browser doesn’t support it” banner. Something like:

```html
<p class="support">
  Your browser doesn't support <a href="TODO">custom CSS functions</a> yet.
</p>
```

```css
@function --supports() {
  result: none;
}

.support {
  padding: 0.5rem;
  display: --supports();
  background-color: #fef0c3;
}
```

There might be a better way of checking support, but I couldn’t find any examples.

css-custom-functions/color-adjust-functions/style.css
chrisdavidmills: OK, that's fair.

css-custom-functions/responsive-narrow-wide/style.css
chrisdavidmills: This is a good tip; I've added it to all three demos in my next commit.

pepelsbey (approved):
Perfect! Thank you :)

PR #329, opened 2025-08-28
Improve compat notes and behavior on popover-hint demo

chrisdavidmills:
As discussed in https://github.com/mdn/content/issues/40881, the browser support situation in our `popover="hint"` demo is not very clear. In browsers that don't support `popover="hint"`, the tooltip behavior is not quite right. And in browsers that don't support CSS anchor positioning, the demo looks broken because the popovers don't appear as expected.

To mitigate these issues, I have updated the demo. Specifically, I have:

- Added more details to the help paragraph about the support situation.
- Moved the popovers inside the `#wrapper` `<div>` and given it a `position` of `relative` so that they should be positioned relative to the `<div>` rather than the `<body>` in non-supporting browsers (note: this still doesan't quite work property. Setting a fallback inset property still seems to position them relative to the `<body>`, and I'm not sure why)
- Updated some of the other code relating to the button bar position to make the above possible.

I'm still not 100% happy with this, but the available information is a lot more useful now, and the demo looks less broken in non-supporitng browsers.

Once we've got this update merged, I'll then look at the associated MDN content and see if any of it needs updating as a result.

### DIFF
--- a/popover-api/popover-hint/index.css
+++ b/popover-api/popover-hint/index.css
@@ -5,7 +5,7 @@
 }
 
 html {
-  font-family: sans-serif;
+  font-family: system-ui;
   height: 100%;
   overflow: hidden;
 }
@@ -14,20 +14,21 @@ body {
   height: inherit;
   background: #ccc;
   margin: 0;
-}
-
-#wrapper {
-  height: inherit;
   display: flex;
   justify-content: center;
   align-items: center;
+}
+
+#wrapper {
+  position: relative;
+  width: 60%;
+  max-width: 500px;
 }
 
 /* Button bar styling */
 
 #button-bar {
   height: 48px;
-  width: 30%;
   min-width: 300px;
   padding: 8px;
   border-radius: 6px;
@@ -36,6 +37,7 @@ body {
   display: flex;
   justify-content: space-around;
   gap: 10px;
+  z-index: 1;
 }
 
 /* Button styling */
@@ -61,6 +63,7 @@ body {
 [popover="auto"] {
   inset: unset;
   position: absolute;
+  bottom: 0;
   bottom: calc(anchor(top) + 20px);
   justify-self: anchor-center;
   margin: 0;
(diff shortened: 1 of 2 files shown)


popover-api/popover-hint/index.css
```diff
@@ -126,11 +129,24 @@ body {
   margin: 0;
   padding: 16px;
   box-shadow: 1px 1px 3px #999;
+  z-index: -1;
```
pepelsbey, line 132: This line makes the paragraph unreachable to the cursor, as it’s hidden behind the body: no text selection, no link clicking.

popover-api/popover-hint/index.html
```diff
@@ -31,32 +31,48 @@
...
-      <button>Option A</button><br /><button>Option B</button>
-    </div>
-    <div id="submenu-2" popover="auto">
-      <button>Option A</button><br /><button>Option B</button>
-    </div>
-    <div id="submenu-3" popover="auto">
-      <button>Option A</button><br /><button>Option B</button>
+      <div id="tooltip-1" class="tooltip" popover="hint">Tooltip A</div>
+      <div id="tooltip-2" class="tooltip" popover="hint">Tooltip B</div>
+      <div id="tooltip-3" class="tooltip" popover="hint">Tooltip C</div>
```
pepelsbey, line 46: I would suggest putting tooltips right next to the buttons they’re triggered by. First of all, this is something you would most likely do in a real-world component to contain everything. Second, it might help to position them next to the buttons as a fallback for browsers that don’t support the `hint` value yet:

```html
<button
  popovertarget="submenu-1"
  popovertargetaction="toggle"
  id="menu-1">
  Menu A
</button>

<div id="tooltip-1" class="tooltip" popover="hint">
  Tooltip A
</div>
```

pepelsbey (commented):
Sorry it took some time! A few suggestions for you to consider :)

pepelsbey (approved):
Great, thank you!

PR #335, opened 2025-09-23
Add notifications demos

wbamberg:
This adds a couple of demos of the Notifications API, one showing persistent notifications, and one showing non-persistent notifications.

I wonder if it might be better to have the persistent one creating the notification in the SW, as well as managing it. That might be a more common use case, but it makes things a little more complicated...

Part of the fix for https://github.com/mdn/content/issues/30931.


chrisdavidmills (approved):
@wbamberg I think this is a really nice set of demos, which shows off the persistent/non-persistent concepts well. I don't have much here except for a few language suggestions, so I'm going to approve it.

PR #339, opened 2025-09-29
Add CSS typed arithmetic examples

chrisdavidmills:
Chrome 140 adds support for [CSS typed arithmetic](https://www.w3.org/TR/css-values-4/#calc-type-checking); see https://chromestatus.com/feature/4740780497043456.

It is not very clear what this is from the spec, but https://css-tricks.com/css-typed-arithmetic/ provides a good explanation.

This PR adds some nice little examples that show what is made possible by this addition.

Note that the linked "Using CSS typed arithmetic" guide does not yet exist; I'm working on that next.


bsmth (approved):
Looks really good, thanks 👍🏻

PR #372, opened 2026-01-08
Initialize year selection to current year

dipikabh:
This is a follow up to https://github.com/mdn/dom-examples/pull/371, which was loading 2030 as the starting year in the form, picking from the first option in the HTML drop-down list.

This PR:
- Sets the year on form load to the current year as well as after a to-do entry.
- Updates the list of years in the drop-down to an ascending order.

### Motivation

To avoid future updates to this UI every time the year changes

### Additional details

The app link can be accessed at https://mdn.github.io/dom-examples/to-do-notifications/

### DIFF
--- a/to-do-notifications/scripts/todo.js
+++ b/to-do-notifications/scripts/todo.js
@@ -42,9 +42,22 @@ window.onload = () => {
     notificationBtn.style.display = "none";
   }
 
-  // Set year to current year on initial load
-  year.value = new Date().getFullYear();
-  note.appendChild(createListItem("App initialised."));
+  // Populate year options starting from the current year
+  // and 12 years into the future
+  const currentYear = new Date().getFullYear();
+  for (let i = 0; i <= 12; i++) {
+    const option = document.createElement("option");
+    const yearValue = currentYear + i;
+    option.value = yearValue;
+    option.textContent = yearValue;
+    year.appendChild(option);
+  }
+
+  // Set the year field to the current year on initial load
+  year.value = currentYear;
+
+  // Log app initialization status
+  note.appendChild(createListItem("App initialized."));
 
   // Let us open our database
   const DBOpenRequest = window.indexedDB.open("toDoList", 4);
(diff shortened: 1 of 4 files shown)


to-do-notifications/scripts/todo.js
```diff
@@ -1,43 +1,61 @@
...
-  const notificationBtn = document.getElementById('enable');
+  const taskList = document.getElementById("task-list");
+  const taskForm = document.getElementById("task-form");
+  const title = document.getElementById("title");
+  const hours = document.getElementById("deadline-hours");
+  const minutes = document.getElementById("deadline-minutes");
+  const day = document.getElementById("deadline-day");
+  const month = document.getElementById("deadline-month");
+  const year = document.getElementById("deadline-year");
+  const notificationBtn = document.getElementById("enable");
```
dipikabh: all automatic linting

to-do-notifications/scripts/todo.js
```diff
@@ -170,45 +212,47 @@ window.onload = () => {
...
-      title.value = '';
+      title.value = "";
       hours.value = null;
       minutes.value = null;
-      day.value = '01';
-      month.value = 'January';
-      year.value = 2026;
+      day.value = "01";
+      month.value = "January";
+      year.value = new Date().getFullYear();
```
dipikabh: I've changed this line (instead of hard-coding it to a year)

pepelsbey (commented):
Looks good! Thank you 😊

I have another idea: we can avoid changing the dates altogether, even in 2030!

This will generate a list of 13 options, starting with the current year (2026–2038)

```js
  // Populate year options: current year and 12 years into the future
  const currentYear = new Date().getFullYear();
  for (let i = 0; i <= 12; i++) {
    const option = document.createElement("option");
    const yearValue = currentYear + i;
    option.value = yearValue;
    option.textContent = yearValue;
    year.appendChild(option);
  }
```

And then we set it:

```js
  year.value = currentYear;
```

I would remove all the pre-filled years from HTML too and leave the `<select>` empty.

I wish I could’ve suggested it, but GitHub won’t allow it 🥲

image

pepelsbey (approved):
Looks good! Thank you 🙂

PR #381, opened 2026-03-15
Fix IndexedDB image preview for uploaded images

### LINKED ISSUE #278: indexeddb-api preview doesn't work on hover

indexeddb-api preview doesn't work on hover

When I successfully uploaded the file, when I hovered over it, I couldn't see the preview information of the photo

I have given the data as follows:

Title: 1
Bibliographic ID: 1
Year: 1
File image: call_of_vahala_img.jpg
Online-file image URL: i dont fill this
(same origin URL)

The photo is displayed in the preview frame

https://mdn.github.io/dom-examples/indexeddb-api/index.html

I have tried many file types like png, jpg, mp4, svg, pdf and the pdf file works.

sbosshardt:
This fixes uploaded-image previews in the indexeddb-api demo.

The image preview code created an object URL, assigned it to the
image, and revoked it immediately. In practice, that could cause
the preview to fail before the image finished loading.

This change revokes the object URL after the image load event
instead, and explicitly loads about:blank in the iframe so the
existing viewer code path runs.

Scope is intentionally limited to images. PDF handling is unchanged.

Fixes #278

### DIFF
--- a/indexeddb-api/main.js
+++ b/indexeddb-api/main.js
@@ -187,9 +187,14 @@
           var img = $('<img id="' + img_id + '"/>');
           $(this).contents().find('body').html(img);
           var obj_url = window.URL.createObjectURL(blob);
-          $(this).contents().find('#' + img_id).attr('src', obj_url);
-          window.URL.revokeObjectURL(obj_url);
+          var imgEl = $(this).contents().find('#' + img_id);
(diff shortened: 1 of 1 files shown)


pepelsbey (approved):
Looks good! Thank you 🙂

PR #382, opened 2026-04-14
Fix message retrieval and variable declaration

philliprobinson:
Replacing let with const for variables where the value is not being mutated downstream.

Returns null if messageBox is not present to begin with.


pepelsbey:
Thank you for trying to help! Unfortunately, I don’t think this fix solves all `null` handling problems in this demo (there are at least two more). But also, it’s not a production code, it’s just a demo. So I would keep it simple. Thanks again!

PR #384, opened 2026-04-29
Use focusNode/Offset instead of extentNode/Offset in EditContext example

ltenmoz:
I have been working on adding EditContext support to Firefox and the EditContext example here has been useful for testing things out. :)
But unfortunately it uses the non-standard `Selection.extentNode`/`Offset` which is not supported on Firefox. Those are now just aliases for the standard `focusNode`/`Offset` see [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/editing/dom_selection.cc;drc=8588b53b02b80dba2d065324bea2ea429bc69b37;l=162) and [here](https://github.com/WebKit/WebKit/blob/1ae0cde7d58ca3346d2b5c4f15ba95701bc36958/Source/WebCore/page/DOMSelection.idl#L78)), so it would be nice to use them instead.

### DIFF
--- a/edit-context/html-editor/converter.js
+++ b/edit-context/html-editor/converter.js
@@ -6,9 +6,9 @@ export function fromSelectionToOffsets(selection, editorEl) {
   const treeWalker = document.createTreeWalker(editorEl, NodeFilter.SHOW_TEXT);
 
   let anchorNodeFound = false;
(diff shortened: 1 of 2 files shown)


pepelsbey (approved):
Good catch! Thank you 🙂

PR #385, opened 2026-06-09
Specify position-anchor: auto in the popover-hint example.

dshin-moz:
Due to back-compat issues [1], popovers don't automatically look for the implicit anchor unless either `position-anchor: auto` is specified, or `position-area` is not `none` (With the default value of `position-anchor: normal`).

This example, because it uses `anchor()` and `anchor-center`, requires explicitly specifying `position-anchor: auto`, as of Bug 2030351.

[1]: https://github.com/w3c/csswg-drafts/issues/13067


keithamus (approved):
We should get this merged, the example looks broken.

PR #386, opened 2026-06-19
Add scroll-promises examples

chrisdavidmills:
In Chrome 150 onwards, the standard `Element` and `Window` scroll methods now return promises. See https://chromestatus.com/feature/5082138340491264. This is particularly useful when a scroll operation is not instantaneous (e.g. you've got `scroll-behavior: smooth;` set on the scrolling container), and you want to programmatically react to the scroll operation finishing, for example, by updating the UI.

Affected methods are:

- `Element.scroll()`
- `Element.scrollBy()`
- `Element.scrollIntoView()`
- `Element.scrollTo()`
- `Window.scroll()`
- `Window.scrollBy()`
- `Window.scrollTo()`

This PR adds a couple of demos that show each promise-returning scroll method in use. I could do these as live examples, but I decided not to, as I don't want to add a huge standalone live example to all seven pages, with all of them doing nearly the same thing. I'd rather provide a couple of code snippets on each page and then link to the full example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,8 @@ Do not specify supported browsers and their versions in code comments or prose, 
 
 - The "screenleft-screentop" directory contains a demo to show how you could use the [Window.screenLeft](https://developer.mozilla.org/docs/Web/API/Window/screenLeft) and [Window.screenTop](https://developer.mozilla.org/docs/Web/API/Window/screenTop) properties to draw a circle on a canvas that always stays in the same physical place on the screen when you move your browser window. [Run the demo live](https://mdn.github.io/dom-examples/screenleft-screentop/).
 
+- The "scroll-promises" directory contains demos illustrating usage of the promise-returning scroll methods (for example, `Element.scroll()`). See the [scroll promises README](scroll-promises/README.md) for more information.
+
 - The "scrolltooptions" directory contains a demo to show how you could use the [ScrollToOptions](https://developer.mozilla.org/docs/Web/API/ScrollToOptions) dictionary along with the [Window.ScrollTo()](https://developer.mozilla.org/docs/Web/API/Window/scrollTo) method to programmatically scroll a web page. [Run the demo live](https://mdn.github.io/dom-examples/scrolltooptions/).
 
 - The "server-sent-events" directory contains a very basic SSE demo that uses PHP to create the server. You can find more information in our [Using server-sent events](https://developer.mozilla.org/docs/Web/API/Server-sent_events/Using_server-sent_events) article. To run the demo you'll need to serve the files from a server that supports PHP; [MAMP](https://www.mamp.info/en/) is a good PHP test server environment.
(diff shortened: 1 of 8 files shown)


hamishwillee (commented):
This looks pretty cool except that after pressing any of these buttons the bar fades and I can't get it back.  I can still press the invisible buttons, but to make the bar reappear I need to refresh the page - not ideal

Image

Edit - this is using latest chrome on refresh, which is still 149

chrisdavidmills:
@hamishwillee darnit, adding the usage of the `interrupted` property from the fulfillment value completely broke it in non-supporting browsers ;-)

To fix this up, I've added a feature detection function and used that to register different event handlers depending on whether the browser supports the promise-returning scroll functions.

I'm then intending to just add a note about the feature detection in the articles, but I thought I'd ask your opinion first.

I'd rather do it this way to avoid making the event handler explanation any more complicated in the articles.

hamishwillee:
I think that makes sense.

scroll-promises/element-methods/index.html
```diff
@@ -0,0 +1,143 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>Scroll promises</title>
```
pepelsbey:
```suggestion
    <title>Scroll promises — Element methods</title>
```

scroll-promises/window-methods/index.html
```diff
@@ -0,0 +1,138 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>Scroll promises</title>
```
pepelsbey:
```suggestion
    <title>Scroll promises — Window methods</title>
```

README.md
```diff
@@ -100,6 +100,8 @@ Do not specify supported browsers and their versions in code comments or prose,
 
 - The "screenleft-screentop" directory contains a demo to show how you could use the [Window.screenLeft](https://developer.mozilla.org/docs/Web/API/Window/screenLeft) and [Window.screenTop](https://developer.mozilla.org/docs/Web/API/Window/screenTop) properties to draw a circle on a canvas that always stays in the same physical place on the screen when you move your browser window. [Run the demo live](https://mdn.github.io/dom-examples/screenleft-screentop/).
 
+- The "scroll-promises" directory contains demos illustrating usage of the promise-returning scroll methods (for example, `Element.scroll()`). See the [scroll promises README](scroll-promises/README.md) for more information.
```
pepelsbey:
```suggestion
- The "scroll-promises" directory contains demos illustrating usage of the promise-returning scroll methods (for example, `Element.scroll()`). See the [scroll promises README](scroll-promises/README.md) for more information.  Run the demos live: [Element methods](https://mdn.github.io/dom-examples/scroll-promises/element-methods/), [Window methods](https://mdn.github.io/dom-examples/scroll-promises/window-methods/).
```

pepelsbey (approved):
Looks good! Thank you 🙂

Just a few suggestions for readmes and titles.

chrisdavidmills:
> Looks good! Thank you 🙂
> 
> Just a few suggestions for readmes and titles.

Super, ta!

PR #388, opened 2026-07-02
Remove outdated comment on "recent browsers"

itsalmostchristmas:
The comment was [added in 2013](https://github.com/mdn/dom-examples/blame/10f6250efb0c6ef1100f777e484ff469270d7d58/to-do-notifications/scripts/todo.js#L83), so is misleading today.


pepelsbey (approved):
Oh, good catch! Thank you 🙂

PR #391, opened 2026-07-15
Remove jQuery from the IndexedDB demo

oiahoon:
Fixes #370

## Summary

- remove the external jQuery 1.8.3 dependency from the IndexedDB demo
- replace its DOM updates, event handlers, and form access with native browser APIs
- preserve add/list/delete/clear behavior and HTML, image, PDF, and fallback blob previews

## Validation

- reproduced the initialization failure with the jQuery CDN request blocked
- verified the complete demo offline: required-field handling, add/list/delete/clear, uploaded HTML/image/fallback previews, and same-origin URL image loading
- verified uploaded HTML renders without executing embedded scripts
- `node --check indexeddb-api/main.js`
- `tidy -errors -quiet indexeddb-api/index.html`
- `git diff --check`


indexeddb-api/main.js
```diff
@@ -4,11 +4,14 @@
...
-    $('#compat-list').append('<li>' + val[0] + ': ' + val[1] + '</li>');
+  var compat = document.querySelector('#compat');
+  var compatList = document.createElement('ul');
+  compatList.id = 'compat-list';
+  compat.replaceChildren(compatList);
+  COMPAT_ENVS.forEach(function(val) {
+    var compatItem = document.createElement('li');
+    compatItem.textContent = val[0] + ': ' + val[1];
+    compatList.append(compatItem);
   });
```
pepelsbey: I don’t think we need that anymore.

pepelsbey (commented):
Thank you for your help! It looks good overall.

I would only suggest making a small step and removing the outdated browser compat part of the script as well. It also goes against the [browser support policy](https://github.com/mdn/dom-examples#browser-support) in the readme.

If you want to go an extra mile, I would say the script still carries some jQuery-specific plumbing that’s not needed for the JS-only version, but it’s not critical. It’s already a big improvement.

PR #392, opened 2026-08-27
Update getUserMedia link in index.html

airbr:
If Im not mistaken, this url I suggest is the actual web tech being used in the code, while the current link goes to the deprecated older version with a bunch of warnings.

I tested the demo in Safari and it works, according to the older deprecated method it shouldn't work as it was removed. Just for completeness.

Thanks for this demo!


pepelsbey (approved):
Oh, good catch! Thank you 🙂

PR #393, opened 2026-09-02
fix: remove incorrect screenLeft polyfill

Josh-Cena:
The example doesn't move the circle if the window starts with `screenLeft=0`, as noticed in https://github.com/mdn/content/pull/45097.

Remove the incorrect feature detection. Every browser supports `screenLeft` anyway, and whenever this fallback runs it assigns static values rather than live aliases, so it is never useful.


dipikabh (approved):
Thank you!

